PR #13, opened 2020-01-19
Implement console_log for sapp-wasm

canadaduane:
Looking for your feedback here. Is this how you'd like to make `console_log` available?

### DIFF
--- a/src/console_log.rs
+++ b/src/console_log.rs
@@ -14,11 +14,11 @@ fn log_record(record: &Record) {
     console_log(&format!("{}", record.args()));
 }
 
-static LOGGER: SappLogger = SappLogger {};
+static LOGGER: WasmLogger = WasmLogger {};
 
-struct SappLogger {}
+struct WasmLogger {}
 
-impl Log for SappLogger {
+impl Log for WasmLogger {
     fn enabled(&self, metadata: &Metadata) -> bool {
         metadata.level() <= log::max_level()
     }
--- a/src/console_log.rs
+++ b/src/console_log.rs
@@ -43,5 +43,5 @@ pub fn init_with_level(level: Level) -> Result<(), SetLoggerError> {
 
 /// Initializes the global logger with `max_log_level` set to `Level::Info` (a sensible default).
 pub fn init() -> Result<(), SetLoggerError> {
-    init_with_level(Level::Info)
+    init_with_level(Level::Debug)
 }
--- a/src/console_log.rs
+++ b/src/console_log.rs
@@ -41,7 +41,7 @@ pub fn init_with_level(level: Level) -> Result<(), SetLoggerError> {
     Ok(())
 }
 
-/// Initializes the global logger with `max_log_level` set to `Level::Info` (a sensible default).
+/// Initializes the global logger with `max_log_level` set to `Level::Debug` (the only supported level for now).
 pub fn init() -> Result<(), SetLoggerError> {
     init_with_level(Level::Debug)
 }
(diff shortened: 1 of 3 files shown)


src/console_log.rs
```diff
@@ -0,0 +1,47 @@
...
+/// Initializes the global logger setting `max_log_level` to the given value.
+pub fn init_with_level(level: Level) -> Result<(), SetLoggerError> {
+    log::set_logger(&LOGGER)?;
+    log::set_max_level(level.to_level_filter());
+    Ok(())
+}
+
+/// Initializes the global logger with `max_log_level` set to `Level::Info` (a sensible default).
+pub fn init() -> Result<(), SetLoggerError> {
+    init_with_level(Level::Info)
```
canadaduane: Oops, will fix this

not-fl3:
Wow, thanks! Thats really nice improvenet - logging on wasm is a big deal! 

My inputs: 

AFAIK rust libraries never embed logging backends. Thats the idea of `log-rs` - that user can choose logging backend implementation. So most idiomatic solution - make separate crate, `sapp-console-log` or something. And let users enable or not that crate in their code. 

But on the other hand - `miniquad` is trying to abstract over platforms and logging is part of that job. 
Filesystem, graphics, even random - are covered by `miniquad`, but logging will require platform dependent `#cfg(...)` and special dependencies for each platform in users code. 
For web - `sapp-console-log` will be actually the only possible logger. And same will be for android/ ios. 

In other words - should we deal with logging in miniquad at all? 

I think we should consider logging as a non-goal for miniquad and do this:

- make separate crate `sapp-console-logger`
- use `sapp-console-logger` in miniquad's user code - more higher level engines. Like in `good-web-game` or `macroquad`
- for `good-web-game` case - add `wasm_log_level` field into the `good_web_game::conf::Conf`. Let it unused on non-wasm platform, but use to choose log level for `sapp-console-logger`
- mention `env-logger` for PC and `sapp-console-logger` for WASM as recommended way to get logs from miniquad. 

Alternatives: make cross platform logging as a part of miniquad. Add feature gates to allow users to override choosen log backend, but make logging works on all platforms by default. The good thing - having `sapp-console-log` crate we can make this decision later.

canadaduane:
What do you think of making other `console.*` log levels available as well? i.e. the way console_log [maps them here](https://github.com/iamcodemaker/console_log/blob/master/src/lib.rs#L121).

canadaduane:
I'm trying to imagine what the code UI for the game developer would look like. Something like this?

```rs
#[macro_use] extern crate log;

#[cfg(target_arch = "wasm32")]
extern crate sapp_console_logger;

#[cfg(not(target_arch = "wasm32"))]
extern crate env_logger;

use log::Level;

fn main() {
    #[cfg(target_arch = "wasm32")]
    sapp_console_logger::init();

    #[cfg(not(target_arch = "wasm32"))]
    env_logger::init();

    debug!("this is a debug {}", "message");
    error!("this is printed by default");

    if log_enabled!(Level::Info) {
        let x = 3 * 4; // expensive computation
        info!("the answer was: {}", x);
    }
}
```

not-fl3:
> I'm trying to imagine what the code UI for the game developer would look like. Something like this?

Yes, exactly. And with ~4 more lines for android/ios. 

I would like to put this code inside `macroquad`. But, actually, for `gwg` maybe it will make sense to keep it outside the library, like it was in [ggez](https://github.com/ggez/ggez/blob/master/examples/logging.rs). Not sure here to be honest.

not-fl3:
> What do you think of making other console.* log levels available as well? i.e. the way console_log maps them here.

Yes, this would be nice! And good reason to split and rename poorly named `test_log` to - separate `console_log`, `console_warn`, `console_error`

PR #14, opened 2020-01-20
Mark sapp-dummy rand unsafe, consistent w/ extern C versions

canadaduane:
Fixes https://github.com/not-fl3/good-web-game/issues/24


not-fl3:
Thanks!

PR #15, opened 2020-01-20
Make `console_*` console log levels available for `sapp-console-log`

canadaduane:
This supercedes #13 and allows for a separate create to take care of `log-rs`-compatibility for wasm32 target.

### DIFF
--- a/native/sapp-wasm/src/lib.rs
+++ b/native/sapp-wasm/src/lib.rs
@@ -252,7 +252,7 @@ pub unsafe fn sapp_run(desc: *const sapp_desc) -> ::std::os::raw::c_int {
             let msg = CString::new(format!("{:?}", info)).unwrap_or_else(|_| {
                 CString::new(format!("MALFORMED ERROR MESSAGE {:?}", info.location())).unwrap()
             });
-            test_log(msg.as_ptr());
+            console_log(msg.as_ptr());
         }));
     }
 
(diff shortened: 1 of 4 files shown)


native/sapp-wasm/src/lib.rs
```diff
@@ -280,17 +280,56 @@ extern "C" {
...
     pub fn canvas_height() -> i32;
-    pub fn test_log(msg: *const ::std::os::raw::c_char);
+    pub fn raw_console_debug(msg: *const ::std::os::raw::c_char);
+    pub fn raw_console_log(msg: *const ::std::os::raw::c_char);
+    pub fn raw_console_info(msg: *const ::std::os::raw::c_char);
+    pub fn raw_console_warn(msg: *const ::std::os::raw::c_char);
+    pub fn raw_console_error(msg: *const ::std::os::raw::c_char);
+}
+
+pub fn console_debug(msg: &str) {
```
not-fl3: I would prefer to keep `sapp_wasm` as low level as possible and build higher level abstractions in other crates. 
Just `fn console_debug(msg: *_)` looks good enough for `sapp_wasm`. 

I think your logging crate is a better place for `console_log(msg: &str)`

native/sapp-wasm/src/lib.rs
```diff
@@ -252,7 +252,7 @@ pub unsafe fn sapp_run(desc: *const sapp_desc) -> ::std::os::raw::c_int {
             let msg = CString::new(format!("{:?}", info)).unwrap_or_else(|_| {
                 CString::new(format!("MALFORMED ERROR MESSAGE {:?}", info.location())).unwrap()
             });
-            test_log(msg.as_ptr());
+            raw_console_log(msg.as_ptr());
```
not-fl3: lets keep it that way until we finish this, but later we can just use `log!()` here, without converting to CString and `#[cfg]`

native/sapp-wasm/src/lib.rs
canadaduane: Oh, I didn't realize `raw_console_*` is available as a public interface. Will change.

canadaduane:
See https://github.com/canadaduane/sapp-console-log for initial implementation of new crate

native/sapp-wasm/src/lib.rs
canadaduane: I don't think `log!()` will be available here, will it? Since we're implementing it at the `sapp-console-log` level, as a dependency of `gwg`?

src/lib.rs
```diff
@@ -21,6 +21,52 @@ use std::ffi::CString;
...
+        panic!("console::warn only implemented for wasm32 target");
+    }
+    
+    pub fn error(msg: &str) {
+        panic!("console::error only implemented for wasm32 target");
+    }
+}
+
+#[cfg(target_arch = "wasm32")]
+pub mod console {
```
not-fl3: Hmmm, console modue is actually a pretty good idea - make platform independent terminal access for all the platforms. 
But maybe it is a bit outside of the scope of this PR?

canadaduane: Ah, I was doing this more to prevent non-wasm32 targets from failing to compile (e.g. when running tests, similar to https://github.com/not-fl3/miniquad/commit/bf56678ed881bf5079befbde121e951f923c7dfa)

It _could_ be used to make it platform independent...

The problem is that `sapp-console-log` doesn't have a way to reach inside of `miniquad` and use the `sapp-wasm` crate... or does it? Can I do `extern crate miniquad::sapp_wasm` from within `sapp-console-log`?

not-fl3:
Perfect!

PR #18, opened 2020-01-26
Focus on canvas to capture keyboard input

canadaduane:
If canvas doesn't get focus, the user must click with mouse before keyboard input is accepted (tested on Firefox).


not-fl3:
Thanks!

PR #21, opened 2020-01-30
OpenGL bindings for macOS

### LINKED ISSUE #4: MacOs support
Build OpenGl version for MacOs and start thinking about IOS afterwards.

akxcv:
Closes #4.

With a ton of help from @not-fl3 himself, I was able to make miniquad work on macOS.

Tested on an early 2015 MacBook Pro running macOS Catalina 10.15.2. Every example works.

Notes:

- Not sure about the necessity of every flag/define I've used.
- The `rand` module for macOS was simply copied from the `sapp-linux` crate, but it appears to work properly.


akxcv:
Not sure if it's possible to build the `x86_64-apple-darwin` version via GitHub Actions... It fails with `unknown option: -fobjc-arc`, which is required for sokol to work.

not-fl3:
Awesome work, thank you!

canadaduane:
This is so great! Thanks @akxcv!

PR #23, opened 2020-02-10
Proof of concept for user-owned Context.

not-fl3:
The issue this PR address: some libraries built on top of miniquad have their own Context structs. It would be nice to store miniquad's context inside those Contexts, without bothering with updating that reference on each event. 

The problem here - in rust its not possibe to have two mutable references to the same data. 

So, if we have `fn mouse_motion_event(..., ctx: &mut Context)` <- this `&mut Context` should be unique. 

Alternatives:
- use Arc and interior mutability. This will require each event handler to receive Arc<Mutex<_>> of Context instead of `&mut Context`, wich is not so nice. Its not possible to build code generic over thread safety here and this will force wasm single threaded users to use Arc/Mutex and will add some boilerplate.

- use interior mutability, but borrow data before event dispatching. This will keep EventHandler signatures the same. But this will make no sense - it will not be possible to use owned clone of Context without UB inside event handlers.

- Deprecate `&mut Context` references for EventHandler. This will solve most of problems, but API with context reference was quite nice.

This PR makes two options for miniquad initialization:

miniquad-owned context:
```
    EventHandler::managed((Stage::new(&mut ctx), ctx))
});
```
Use will get Context, but could move it back to miniquad and later receive in EventHandlder as a reference. Stage will implement old `EventHandler` trait with `&mut Context` argument in each function.

user-owned context:
```
miniquad::start(conf::Conf::default(), |ctx| EventHandler::owned(Stage::new(ctx)));
```
User consume ctx. `Stage` should implement new `EventHandler` trait without `&mut Context` argument.

This safety and correctness is guaranteed on type level, its impossible to implement wrong trait or occasionally store wrong reference. 
But implementation is not really nice. 
Will think a bit about all this. Maybe will just clean up naming in implementation and merge.

### DIFF
--- a/examples/instancing.rs
+++ b/examples/instancing.rs
@@ -141,7 +141,9 @@ impl EventHandler for Stage {
 }
 
 fn main() {
-    miniquad::start(conf::Conf::default(), |ctx| Box::new(Stage::new(ctx)));
+    miniquad::start(conf::Conf::default(), |mut ctx| {
+        UserData::owning(Stage::new(&mut ctx), ctx)
+    });
 }
 
 mod shader {
--- a/examples/offscreen.rs
+++ b/examples/offscreen.rs
@@ -200,7 +200,7 @@ impl EventHandler for Stage {
 }
 
 fn main() {
-    miniquad::start(conf::Conf::default(), |ctx| Box::new(Stage::new(ctx)));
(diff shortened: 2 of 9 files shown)


not-fl3:
Update: tried on couple of examples and macroquad. I think that naming is terrible and very confusing, but the idea is good, I like it.

Issue #26, opened 2020-02-20
Mouse relative movement


not-fl3:
The problem: it is impossible now to implement infinite mouse movements wich miniquad, wich is highly usefull for games.

This involves two major components:

- [ ] grab mouse input to miniquad's window
- [ ] receive dx/dy in "mouse_motion_event" event

See a similar issue in godot repo with nice gifs explaining different possible grabbing modes: https://github.com/godotengine/godot/issues/21751

not-fl3:
Fixed!

PR #27, opened 2020-02-21
Mouse capture and relative movement

not-fl3:
Related issue: https://github.com/not-fl3/miniquad/issues/26 

It works for linux and wasm
But for linux it required linking with an additional library: libXi - library for the X Input Extension. 

Need to think about how bad it is. If libXi is not preinstalled on Ubuntu - it will be a problem.

### DIFF
--- a/native/sapp-darwin/src/sokol_app_darwin.rs
+++ b/native/sapp-darwin/src/sokol_app_darwin.rs
@@ -47,7 +47,8 @@ pub const sapp_event_type_SAPP_EVENTTYPE_SUSPENDED: sapp_event_type = 17;
 pub const sapp_event_type_SAPP_EVENTTYPE_RESUMED: sapp_event_type = 18;
 pub const sapp_event_type_SAPP_EVENTTYPE_UPDATE_CURSOR: sapp_event_type = 19;
 pub const sapp_event_type_SAPP_EVENTTYPE_QUIT_REQUESTED: sapp_event_type = 20;
-pub const sapp_event_type__SAPP_EVENTTYPE_NUM: sapp_event_type = 21;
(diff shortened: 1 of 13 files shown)


koalefant:
Thanks for adding this! :heart: 

I have tested two scenarios so far:
1. Click-and-drag, like map scrolling in RTS. This works well in both Chrome and Firefox for me :+1: On Linux, unlike in Browser, mouse-move does not get inhibited when raw-mouse arrives. So currently one would need to test build target to workaround this. It could also be nice to hide cursor for consistency on Linux.

2. Mouse-look, without holding a button - works fine in the browser (as long as activated from mouse press handler). On Linux there is another issue - mouse is still restricted to the game window even when it loses focus. Perhaps focus deactivation could be tracked by the library?

koalefant:
Another weird issue on Linux: units appear to be different from pixels, which results in different motion speed in browser/desktop?

not-fl3:
> It could also be nice to hide cursor for consistency on Linux.

Fixed,  now it is possible to hide mouse cursor on desktop `ctx.show_mouse(false)`:
grab

> Another weird issue on Linux: units appear to be different from pixels, which results in different motion speed in the browser/on the desktop.

hmm interesting! but I guess this is how X11 works - raw events are in some other units and there is no easy fix. But need some research here, thanks for the report!

> mouse-move does not get inhibited when raw-mouse arrives

that's true, but I don't have any ideas on how to do it better right now. I think it would be nice to stop mouse_move events with the captured cursor on web,  any other ideas would be appreciated! 

> mouse is still restricted to the game window even when it loses focus

Yes, also true. Mouse grab is not automatically released by X11 when focus lost and AFAIK most common solution - the program should catch focus events and release the cursor on focus lost. 
Focus events are not yet implemented: https://github.com/not-fl3/miniquad/issues/28

PR #34, opened 2020-03-03
Fix link to macroquad from Readme

Vrixyz:
`github.com/not-fl3/macroquad` actually redirects to `https://github.com/not-fl3/miniquad/blob/master/github.com/not-fl3/macroquad` which is a wrong link


not-fl3:
Thanks!

PR #35, opened 2020-03-04
Pipeline::with_params(): mention missing attribute name when glGetAttribLocation fails

koalefant:
helps debugging while writing shaders


not-fl3:
Nice, thanks!

PR #42, opened 2020-03-16
Emulate mouse move with touch on phone

nokola:


nokola:
I was testing miniquad on my phone and noticed the touch events were not coming through. I noticed it's because touch_event was not forwarding the touch-move-s to the mouse. Not sure if it was intentional or not. Either way - here's the change to fix it in case it wasn't intentional.

not-fl3:
Nice, thank you!
 
I think I need to make a wiki page describing input handling with all corner cases on different platforms and fix the decisions on events emulation there.

PR #45, opened 2020-03-25
Pipeline::with_parms(): added assertion to catch vertex attribute stride above 255 ASAP (WebGL 1 limitation)

koalefant:
I imagine it is nicer to be able to catch this in native build instead of debugging WebGL-one


not-fl3:
Nice, more asserts are always good, thanks!

PR #46, opened 2020-03-25
load_shader: do not escape shader error_message as it makes it less readable

koalefant:
Removes escaping from shader compiler errors (introduced by debug formatting {:?}).


not-fl3:
Thanks!

PR #48, opened 2020-03-29
Support more GL functions (enough to implement 2d vector library like with NanoVG)

nokola:
# Summary 

I ported significant part of [NanoVG (nvg-rs)](https://github.com/sunli829/nvg) over to miniquad. NanoVG is high-quality antialiased drawing and text library. I think it's the first time I see it run on WASM, Desktop, and Android from rust! 

I created new library and named it nvg-miniquad (not yet published). I plan to publish the **nvg-miniquad** library, but first I want to make sure you approve of the changes to **miniquad**.

I made these changes in miniquad to enable support for libraries like nvg-miniquad:
1. Found some existing GL functions were crashing on WASM and missing on Windwows, added those
2. Changed type of GL_FALSE and GL_TRUE to u8, otherwise I got lots of errors trying to pass the u32 to some gl* functions that need u8. Now they can be passed to gl functions using u32 or u8.
3. Exported Sokol's gl* functions for cross-platform full access to GL! (Thanks for making the xplat support in miniquad, it's very good.)

# Example  

## Windows
image

## WebGL
image

## Sample code
Here is partial example code to render the above from the nvg-miniquad ported library:

```rust
        self.nvg_context.begin_path();
        self.nvg_context.rect((100.0, 100.0, 300.0, 300.0));
        self.nvg_context.fill_paint(nvg::Gradient::Linear {
            start: (100, 100).into(),
            end: (400, 400).into(),
            start_color: nvg::Color::rgb_i(0xAA, 0x6C, 0x39),
            end_color: nvg::Color::rgb_i(0x88, 0x2D, 0x60),
        });
        self.nvg_context.fill().unwrap();

        let origin = (150.0, 140.0);
        self.nvg_context.begin_path();
        self.nvg_context.circle(origin, 64.0);
        self.nvg_context.move_to(origin);
        self.nvg_context.line_to((origin.0 + 300.0, origin.1 - 50.0));
        self.nvg_context.stroke_paint(nvg::Color::rgba(1.0, 1.0, 0.0, 1.0));
        self.nvg_context.stroke_width(3.0);
        self.nvg_context.stroke().unwrap();

        self.nvg_context.end_frame().unwrap(); // comment to show stuff
```

Note: I haven't tested text and there are very likely issues still with blend modes, however shapes and fill look OK right now! :)

Any feedback welcome! Thanks for making miniquad very easy to be cross-platform and run on mobile as well as WASM/Desktop again!


not-fl3:
Wow, great PR!
Always nice to see miniquad in action!

nokola:
Thanks! I had to spend some time to update the original nvg port to make sure everything works on Android/WASM/Desktop :)

native/sapp-wasm/src/gl.rs
```diff
@@ -739,8 +739,8 @@ extern "C" {
 extern "C" {
     pub fn glClearStencil(s: GLint);
 }
-extern "C" {
-    pub fn glColorMask(red: GLboolean, green: GLboolean, blue: GLboolean, alpha: GLboolean);
+pub fn glColorMask(red: GLboolean, green: GLboolean, blue: GLboolean, alpha: GLboolean) {
+    // TODO: marking this extern "C" does not work on WASM,  shows: Uncaught (in promise) LinkError: WebAssembly.instantiate(): Import #46 module="env" function="glColorMask" error: function import requires a callable
```
not-fl3: I am looking on colorMask and stencil function in docs and it looks that they should be supported in webgl1: 

https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/colorMask
https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/stencilOp

The other clue that they are probably supported: there is no special code in orignal emscripten js code: https://github.com/emscripten-core/emscripten/blob/master/src/library_webgl.js (just search for appropriate function, its a long file)

So my guess is that if you add all that function to gl.js - they will work in webgl. 

Something like this will fix wasm runtime errors and I guess will work on webgl. But hard to say without tests, I am never sure when its about web
```
--- a/native/sapp-wasm/js/gl.js
+++ b/native/sapp-wasm/js/gl.js
@@ -431,6 +431,9 @@ var importObject = {
         glClearStencil: function (s) {
             gl.clearColorStencil(s);
         },
+        glColorMask: function (red, green, blue, alpha) {
+            gl.colorMask(red, green, blue, alpha);
+        },
         glScissor: function (x, y, w, h) {
             gl.scissor(x, y, w, h);
         },
```

nokola: Thanks - will try the proposed fix!

native/sapp-windows/src/sokol_app_msvc.rs
```diff
@@ -42386,6 +42395,19 @@ extern "C" {
 extern "C" {
     pub fn glBindBuffer(target: GLenum, buffer: GLuint);
 }
+
+// TODO: make this work by linking to library that supports it
+// extern "C" {
+//     pub fn glBindBufferRange(
```
not-fl3: from the docs - glBindBufferRange should be available on OpenGl 3.3
but it is not loaded in the C code - because it was not used in miniquad

if it is the only issue - you can try this fix(did not really checked, so its a blind fix): 

```
diff --git a/native/sapp-windows/external/sokol/sokol_app.h b/native/sapp-windows/external/sokol/sokol_app.h
index 4389a6e..a1ca5ff 100644
--- a/native/sapp-windows/external/sokol/sokol_app.h
+++ b/native/sapp-windows/external/sokol/sokol_app.h
@@ -3375,6 +3375,12 @@ void glBindBuffer(GLenum target, GLuint buffer) {
     _sapp_glBindBuffer(target, buffer);
 }
 
+typedef void  (GL_APIENTRY *PFN_glBindBufferRange)(GLenum target, GLuint index, GLuint buffer, GLintptr offset, GLsizeiptr size);
+static PFN_glBindBuffer _sapp_glBindBufferRange;
+void glBindBufferRange(GLenum target, GLuint index, GLuint buffer, GLintptr offset, GLsizeiptr size) {
+    _sapp_glBindBufferRange(target, index, buffer, offset, size);
+}
+
 typedef void  (GL_APIENTRY *PFN_glDeleteVertexArrays)(GLsizei n, const GLuint * arrays);
 static PFN_glDeleteVertexArrays _sapp_glDeleteVertexArrays;
 typedef void  (GL_APIENTRY *PFN_glDepthMask)(GLboolean flag);
@@ -3579,6 +3585,7 @@ _SOKOL_PRIVATE  void _sapp_win32_gl_loadfuncs(void) {
     _SAPP_GLPROC(glDisable);
     _SAPP_GLPROC(glColorMask);
     _SAPP_GLPROC(glBindBuffer);
+    _SAPP_GLPROC(glBindBufferRange);
     _SAPP_GLPROC(glDeleteVertexArrays);
     _SAPP_GLPROC(glDepthMask);
     _SAPP_GLPROC(glDrawArraysInstanced);
```

src/lib.rs
```diff
@@ -1,21 +1,21 @@
...
+pub extern crate sapp_darwin as sapp;
 #[cfg(not(any(
     target_os = "linux",
     target_os = "macos",
     target_os = "android",
     target_arch = "wasm32",
     windows
 )))]
-extern crate sapp_dummy as sapp;
+pub extern crate sapp_dummy as sapp;
```
not-fl3: I am seriously concerned about exposing internal gl functions to users. 

I understand that this may be helpful for platform-dependent extensions for miniquad, but there are a couple of problems:
- miniquad use a lot of internal caches and it is really easy to break gl state by raw gl calls
- sapp-* crates are not designed as a general purpose GL loaders 
- both problems will be even more serious with multiple graphics backends in the future

nokola: hmm, that's a good point. Do you mean for example that Metal support will not work? Would the solution be to update Context to support everything nanovg uses in such case? I like having the GL calls, because I can iterate on my graphics library quickly, but I see the good option of being more platform agnostic.

not-fl3: If it is possible - it is better to expose those functions from the context for sure. 
Is it really much of things missing? Maybe I can help with some of those things? 

I can give an example - ColorMask should belong to miniquad's Pipeline. 
And it even is in PipelineParams - as `color_write`. But the actual call to glColorMask was unimplemented.
I guess it is the same for other functions you had to expose from gl.

not-fl3: WIP `color_mask` PR: https://github.com/not-fl3/miniquad/pull/49 :)

nokola: Nice! I will look again at the source and make a list of things missing in Context, and see how long will it take to move them in Context.

nokola: For this PR, I propose to remove the public crate sapp::* and still keep the other benefits. Will make an update to the PR and push today/tomorrow.

nokola:
Note: I published https://github.com/nokola/nvg-miniquad so you can see what calls it is currently using. My intention is to try and move nvg-miniquad over to use the Context instead of GL* calls as discussed.

not-fl3:
Awesome, thanks! 
I just merged the glColorMask PR and going to try to run `nvg-miniquad` and implement missing features.
Let me know if you are working on some function to not overlap the work.

not-fl3:
Oh, I see, `nvg-miniquad` is using miniquad as a gl loader and is implemented in raw opengl. In contrast typical miniquad's code looks like: https://github.com/not-fl3/miniquad/blob/master/examples/quad.rs

My guess is that you need a cross-platform opengl loader, and this was not a goal for miniquad. Probably something like this can be built with slightly modified sapp-* crates thought

nokola:
I am really looking for cross-platform 2d antialiasing drawing library. Whether it's raw opengl or something else - either will work. If I build with opengl loader, would it be useful later with iOS/metal? Also I like miniquad's cross-platform code and would like to reuse it, especially if it helps support iOS later. What do you think is the best way to proceed? I can try switching to miniquad's Context and not use raw opengl, or would opengl be better in your opinion? I'm thinking if we can build a good antialiasing library on top of miniquad's Context might be best.

not-fl3:
I would recommend trying miniquad's API - with Buffers, Pipelines, Bindings etc. This API eventually will be implemented on top of different graphics API - starting from Metal. 
I personally really like that level of abstraction - I used it a lot back in C++ days with Oryol/Sokol and really enjoyed it. 

However, it is really exciting that miniquad can provide cross platform opengl context while was not designed to do it :) Maybe we should somehow make it available for those who just want to use raw graphics APIs.

nokola:
> However, it is really exciting that miniquad can provide cross platform opengl context while was not designed to do it :) Maybe we should somehow make it available for those who just want to use raw graphics APIs.

Yes, I agree! Having the opengl context for experimenting is exciting! 
What would be the best way to do it? How about another library e.g. miniquad-gl that lives in the same repository as miniquad? The nvg Rust port is implemented in a similar way - there's a root folder nvg library https://github.com/sunli829/nvg and nvg-gl (which is the equivalent of miniquad-gl) living in the same nvg repository as subfolder: https://github.com/sunli829/nvg/tree/master/nvg-gl

One potential structure could be:
```
miniquad/src -> stays the same as now
miniquad/miniquad-gl -> new library as subfolder that provides cross platform opengl context
miniquad/miniquad-gl/native is the current contents of miniquad/native
```
Either way, I'll try to move the nvg-minuquad to use Context instead of gl calls directly, because I'd like to eventually gain free support for Metal when miniquad supports it.

not-fl3:
If you are going to try using miniquad's rendering api - I am looking forward to your feedback! 

I personally found it way better for both prototyping and development as a raw opengl.

not-fl3:
For the cross platform raw gl library - I would consider doing it as a separate library. The goals of general-purpose opengl loader are a bit different - support different opengl versions, cover the whole opengl api etc

nokola:
Yes, good point. Also I like the simplicity of the current miniquad and simple/no dependencies.

native/sapp-windows/src/sokol_app_msvc.rs
nokola: Updated, thanks for the pointer and the code! I fixed both bindbufferrange and stencilopseparate - no more TODOs  :)

nokola:
All ready to review/merge! Verified nvg-miniquad works on WASM, Desktop, Android still. Thanks for the help!

nokola:
@not-fl3: All ready to review/merge! Verified nvg-miniquad works on WASM, Desktop, Android.
Thanks for the help!

not-fl3:
I still can't see miniquad as a general-purpose opengl context loader library.  Maybe I will find the ways to do this in the future though or just missing some good ways to do it.

Right now the only idea I have for your use case - remove all the rendering part from your version of miniquad and release it as a different library with a different name.

nokola:
> I still can't see miniquad as a general-purpose opengl context loader library.

I agree - check out the files I reverted and what is left - there's no longer sapp::* public and no changes that change the idea of miniquad. 
All the changes are now internal and can be used to improve miniquad's Context and I think not towards general library.

nokola:
If I missed something in your intention please let me know! I think after the stuff I took out, the PR is aligned with the intention of miniquad.
My goal is to build the nvg-miniquad on top of Context, and maybe do some updates/etc in Context to support it (future PR)

not-fl3:
Oh, I see, `sapp-*` crates are available as separate crates and may be imported directly (or through `miniquad::sapp`, wich is the same).  
Nice! This way low-level API is available but is not miniquad's responsibility anymore, I like it!

I still don't quite understand what exactly do you want to get - do you want to use miniquad or opengl as a rendering API? If your goal is to use opengl - ok, I think I am fine with directly using `sapp-*` crates. In this case, I will review this PR and will eventually merge it. 

If you are going to eventually switch to miniquad - I would better use this PR as a base for the future PRs implementing missing miniquad's features, but would not merge it right now.

nokola:
> I still don't quite understand what exactly do you want to get - do you want to use miniquad or opengl as a rendering API?
I want to use miniquad as rendering API instead of OpenGL (which I'm currently using.)
Basically, I want to change my [nvgimpl.rs](https://github.com/nokola/nvg-miniquad/blob/master/src/nvgimpl.rs) to only use miniquad::Context and no direct OpenGL calls. 

I see this as 3 step process:
1. Add low-level OpenGL functions to miniquad::sappXX that will be needed in step 2 (**done by current PR**)
2. Add minimal set of features to miniquad::Context - enough to support the needs of nvg-miniquad (future PR)
3. Update nvg-miniquad to use miniquad::Context instead of OpenGL calls.

I was playing with the low-level sapp::* APIs for the last few days to see if I can port nanovg, but ideally I'd like to move away from them in a way that can benefit both users of miniquad who want 2D vector library and my cross-platform vector app that I'm building.

Hope that clarifies it - please let me know!
I can also help with iOS support in miniquad (I have a Mac and iPhone) at some time in the future.

> Nice! This way low-level API is available but is not miniquad's responsibility anymore, I like it!

Just a note: I think they may be importable separately, but not through miniquad::sapp, because miniquad::sapp is private.

not-fl3:
> I think they may be importable separately, but not through miniquad::sapp, because miniquad::sapp is private. 

I also thought that way, but apparently `use miniquad::sapp` works even without `pub extern crate`.
Wich is weird for me, but makes some sense - it would be possible to use it anyway with direct use. 

> I want to use miniquad as rendering API instead of OpenGL (which I'm currently using.)

Ok! Let's postpone this PR then and add all the missing features one by one, using code from this PR when those OpenGl functions will be needed. 
From the functions list, I guess there are going to be one stencil-buffer PR similar to color mask PR https://github.com/not-fl3/miniquad/pull/49

and maybe some change on the way miniquad update buffers. 

Let me know how it goes with the port to miniquad!

nokola:
Sounds good! I'll close this PR for now, make the Context changes in my nokola/miniquad fork either one by one or all at once (depending on how much time it takes) 
Then I'll PR the OpenGL+Context changes to miniquad either all at once or separately, again depending on how much time it takes to do each.

PR #49, opened 2020-03-29
Implement color mask

not-fl3:
Actually use `color_write` field of PipelineParams for masking pipeline channels


not-fl3:
colormask

it works! :)

PR #51, opened 2020-03-30
Support UniformType::Int

nokola:


src/graphics.rs
```diff
@@ -39,6 +40,7 @@ impl UniformType {
             UniformType::Float2 => 8 * count,
             UniformType::Float3 => 12 * count,
             UniformType::Float4 => 16 * count,
+            UniformType::Int => 4 * count,
```
not-fl3: If it is possible - maybe add Int1, Int2, Int3 and Int4 as well?

nokola: done

not-fl3:
Great, thanks!

PR #57, opened 2020-04-08
Fix support for glUniformXiv

nokola:


nokola:
Verified `cargo run --example quad` and `cargo run --example offscreen` works on Windows-MSVC

not-fl3:
Thank you! Merging now, let's see what the new CI will think about this :)

PR #60, opened 2020-04-08
Update rust.yml

AKholetsky:
set fail-fast to false to proceed other tasks when one of build failed


AKholetsky:
[Fail-fast docs](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast)

not-fl3:
Thanks!

PR #66, opened 2020-04-18
Deprecate "miniquad::rand"

not-fl3:
"quad-rand" is new recommended way to get random numbers in miniquad. 

Using rand from libc causes too many problems and is hard to maintain.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -32,7 +32,18 @@ pub use graphics::*;
 
 use std::ffi::CString;
 
-pub use sapp::{rand, RAND_MAX};
+#[deprecated(
+    since = "0.3",
+    note = "libc rand is slow and incosistent across platforms. Please use quad-rnd crate instead."
+)]
+pub unsafe fn rand() -> i32 {
+    sapp::rand()
+}
+#[deprecated(
+    since = "0.3",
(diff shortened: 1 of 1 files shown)


nokola:
btw, there are few other pcg random number libraries available: https://docs.rs/pcg/3.0.1/pcg/ and https://docs.rs/pcg_rand/0.11.1/pcg_rand/
Should we use one of those instead? I haven't looked at them in detail, just asking.

not-fl3:
```
    let mut pcg = Pcg32::from_entropy();

    let x : u32 = pcg.gen();
```

you will need to store `pcg` somewhere and pass it over to each function using random. And this is basically the same for all other prng libs available. 
While `quad-rand` uses static atomics internally and is very similar to libc's rand. 

Features unique to quad-rand: 
- no need to store generator
- nice helper functions like `gen_range(min, max)` for most standard type, Vec::shuffler, Vec::choose, Vec::choose_multiple
- optional compatibility layer with `rand` crate - it is possible to use `rand`'s statistics stuff with `quad-rand` random generator

For better control it is possible to use any other random crate for sure, `quad-rand` is just recommended one.

nokola:
Thanks for the info, good to know!

PR #67, opened 2020-04-20
Support creating textures without data

nokola:
Helps to allocate GPU textures without data transfer - can be used to reduce memory fragmentation for UI/vector tools


src/graphics/texture.rs
```diff
@@ -189,8 +194,10 @@ impl Texture {
     }
 
     /// Upload RGBA8 texture to GPU
-    pub fn from_rgba8(ctx: &mut Context, width: u16, height: u16, bytes: &[u8]) -> Texture {
-        assert_eq!(width as usize * height as usize * 4, bytes.len());
+    pub fn from_rgba8(ctx: &mut Context, width: u16, height: u16, bytes: Option<&[u8]>) -> Texture {
```
not-fl3, line 197: I guess `fn from_rgba8<T: Into<Option<&[u8]>>>(..)` will allow to do the same and keep api-compatibility with previous versions as well?

not-fl3:
I agree, its a good idea!

src/graphics/texture.rs
nokola: checking!

nokola: I couldn't get the `Into<...>` solution to work without breaking existing code, because of the following error. See this playground: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f828db0211bf78fea4b34b1da34138e8
sample code from playground:
```rust
fn main() {
    println!("Hello, world!");
    let pixels: [u8; 4 * 4 * 4] = [
        0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0x00, 0x00, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0x00,
        0x00, 0xFF, 0xFF, 0x00, 0x00, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0x00, 0x00, 0xFF,
        0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0x00, 0x00, 0x00, 0xFF, 0xFF,
        0xFF, 0xFF, 0xFF, 0x00, 0x00, 0xFF, 0xFF, 0x00, 0x00, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
        0xFF, 0x00, 0x00, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
    ];
    process_pixels(&pixels);
    process_pixels_optional(&pixels);
}

fn process_pixels(bytes: &[u8]) {
    println!("pixels");
}

fn process_pixels_optional<'a, T: Into<Option<&'a[u8]>>>(bytes: T) {
    println!("pixels");
}
```

Error:
```log
   Compiling playground v0.0.1 (/playground)
error[E0277]: the trait bound `std::option::Option<&[u8]>: std::convert::From<&[u8; 64]>` is not satisfied
  --> src/main.rs:11:5
   |
11 |     process_pixels_optional(&pixels);
   |     ^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::convert::From<&[u8; 64]>` is not implemented for `std::option::Option<&[u8]>`
...
18 | fn process_pixels_optional<'a, T: Into<Option<&'a[u8]>>>(bytes: T) {
   |    -----------------------        --------------------- required by this bound in `process_pixels_optional`
   |
   = help: the following implementations were found:
             <std::option::Option<&'a T> as std::convert::From<&'a std::option::Option<T>>>
             <std::option::Option<&'a mut T> as std::convert::From<&'a mut std::option::Option<T>>>
             <std::option::Option<T> as std::convert::From<T>>
   = note: required because of the requirements on the impl of `std::convert::Into<std::option::Option<&[u8]>>` for `&[u8; 64]`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0277`.
error: could not compile `playground`.

To learn more, run the command again with --verbose.
```

To workaround, one can call `process_pixels_optional(&pixels[..]);` (notice the added [..]), however I think adding [..] is not intuitive. 

After thinking about it some more, I think it would be best to break existing code, but have the current PR that clearly uses Some(...) or None, because it makes it easier to discover that textures can be created without pixels and it feels more idiomatic.

Wondering what do you think about the subject.

not-fl3: Oh, that is true. We can rewrite require writing `from_rgba8(&colors[..])` instead `from_rgba8(&colors)`, but I dont really like it - the error is too confusing and frustrating, and `from_rgba8` signature with lifetimes looks a bit scary.

Meanwhile, I got the new argument - semantically `from_rgba8` function should accept an array of colors, not option.

So I propose leaving `from_rgba8` without Option and modify only `from_data_and_format` instead. This makes sense for me - from_rgba8 wants colors, but when you need texture with specific format but no data - go ahead and use `from_data_and_format`.

What do you think?

nokola: > Meanwhile, I got the new argument - semantically `from_rgba8` function should accept an array of colors, not option.

I meant: both `from_rgba8` and `from_data_and_format` should accept `Option<...>`:
```
from_rgba8(..., Some(&colors));
from_data_and_format(..., Some(&colors));
```
The new argument is: The "Some(...)" is helpful in both cases because semantically you can either pass something to the texture, or nothing in both cases and it feels more like idiomatic Rust - things are explicit not implicit. I think both functions should have `Option<...>` because from API surface level they both feel similar (`from_xx` functions.) If I use from_data_and_format(...) with Option I would expect to also use from_somethingelsehere(...) with Option as well.

not-fl3: I would not really mind Option in `from_rgba8` if it would be possible without breaking change.
With breaking change - I think it is not worse it.
And the argument - `from_rgba8` creates a texture from pixels. If you need something else - go ahead and use `from_data_and_format` and create whatever is needed.

not-fl3: However I am open to break the api, especially in 0.3, so if you found it really important to make it Option - I am fine with that and will just merge this.

not-fl3: In sokol-gfx there is just one function, kind of similar to our `from_format_and_data` with huge params struct.
All the fields are null by default and null in data is supported in a similar way that you did in `from_format_and_data`.

not-fl3: In sokol-gfx there is just one function, kind of similar to our `from_format_and_data` with huge params struct.
All the fields are null by default and null in data is supported in a similar way that you did in `from_format_and_data`.

nokola:
Created new PR https://github.com/not-fl3/miniquad/pull/70 - open to suggestions, thanks!

PR #68, opened 2020-04-20
Support Alpha textures

nokola:


not-fl3:
The problem here - on WebGL 1 there are only GL_ALPHA for single-channel textures, where that single channel is ALPHA, while on ES3/gl3 its only R8, where that channel is in ALPHA. 

Miniquad had R8 textures before, but I decided that it's better to introduce some compatibilities-checking mechanism before adding platform-dependent features.

nokola:
Hmm, interesting! Does it mean the pixel shader code will also have to be different on webgl1 vs gl3, or can work on both if the pixel shader code references ".a"?
If you have suggestions for what the compatibility checking would be, I can try adding it.

not-fl3:
That's a problem that I failed to solve - R8 textures did not put data into the alpha channels, so shaders should be different. 

I think we can merge it with some comments on limited platform support, but I would rather postpone it. 

From compatibilities I want something like: 
```
if ctx.compatibilities.texture_format_supported(TextureFormat::R8) {
}
```
in some generalized way for all platform-dependent features. But I did not even started to properly think about it.

nokola:
Yes - compatibility is an interesting case. For the Alpha texture support - in the future maybe the easy way is to say "only support opengl es3+ and webgl 2+" 

Although for sure compatibility will be useful at some point.

Side info: Interesting link I found for cross-compilation of shaders: https://github.com/Unity-Technologies/HLSLcc

PR #70, opened 2020-04-21
Support texture without data - extra Texture::new function

nokola:
Cleaned up code and created new PR

### DIFF
--- a/src/graphics/texture.rs
+++ b/src/graphics/texture.rs
@@ -121,7 +121,7 @@ impl Texture {
         Self::new(ctx, TextureAccess::RenderTarget, None, params)
     }
 
-    pub fn new(ctx: &mut Context, access: TextureAccess, bytes: Option<&[u8]>, params: TextureParams) -> Texture {
+    pub fn new(ctx: &mut Context, _access: TextureAccess, bytes: Option<&[u8]>, params: TextureParams) -> Texture {
         if let Some(bytes_data) = bytes {
             assert_eq!(
                 params.format.size(params.width, params.height),
@@ -137,9 +137,6 @@ impl Texture {
 
         unsafe {
             glGenTextures(1, &mut texture as *mut _);
-            if access == TextureAccess::RenderTarget {
-                glActiveTexture(GL_TEXTURE0);
-            }
             ctx.cache.bind_texture(0, texture);
             glTexImage2D(
                 GL_TEXTURE_2D,
--- a/src/graphics/texture.rs
+++ b/src/graphics/texture.rs
@@ -97,6 +97,15 @@ pub enum FilterMode {
     Nearest = GL_NEAREST as isize,
 }
 
+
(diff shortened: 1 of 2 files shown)


nokola:
Verified these examples work on Windows:
1. quad
2. offscreen
3. instancing

src/graphics/texture.rs
```diff
@@ -116,7 +137,9 @@ impl Texture {
 
         unsafe {
             glGenTextures(1, &mut texture as *mut _);
-            glActiveTexture(GL_TEXTURE0);
+            if access == TextureAccess::RenderTarget {
+                glActiveTexture(GL_TEXTURE0);
```
not-fl3, line 141: Why only for RenderTarget textures?

nokola: I am not sure - I copied the logic in `from_data_and_format` below - the only difference between `from_data_and_format` and `new_render_texture` was that `from_data_and_format` does not call glActiveTexture(...).

src/graphics/texture.rs
```diff
@@ -147,45 +173,7 @@ impl Texture {
...
-        );
-
-        let (internal_format, format, pixel_type) = params.format.into();
-
-        unsafe {
-            ctx.cache.store_texture_binding(0);
-
-            let mut texture: GLuint = 0;
-            glGenTextures(1, &mut texture as *mut _);
-            ctx.cache.bind_texture(0, texture);
```
nokola, line 162: just for reference: this is the place where glActiveTexture() was missing - not sure if by accident or intentional

src/graphics/texture.rs
not-fl3: It looks like calling glActiveTexture always is fine. At least I cant see anything broken without that `if` :)

not-fl3: Oh wait, 
```
    fn bind_texture(&mut self, slot_index: usize, texture: GLuint) {
        unsafe {
            glActiveTexture(GL_TEXTURE0 + slot_index as GLuint);
      ...
```

glActiveTexture is going to be called anyway, glActiveTexture should be just deleted!

not-fl3:
Awesome, thanks!

src/graphics/texture.rs
nokola: nice! Removed. I still kept the TextureAccess even though unused. Do you think we'll need TextureAccess later? 

I know for example Unity differentiates RenderTexture (render target) vs "regular" Texture2D (GPU read), however in the miniquad code there's currently no implementation difference.
I assume the TextureAccess can be used to speed up _something_ but I'm not sure what - perhaps in the future some optimization with framebuffer object vs texture? (https://community.khronos.org/t/using-textures-as-render-target/66561)

not-fl3: Yes, I am pretty sure that there are some possible optimizations with TextureAccess, so its a good change!

Issue #55, opened 2020-03-31
Support for pixel arithmetic for RGB and alpha components separately


nokola:
The BlendState today only supports the equivalent of BlendFactor for argb (glBlendFunc):
```rust
type BlendState = Option<(Equation, BlendFactor, BlendFactor)>;
```

Ideally, it would support separate RGB and alpha blends (equivalent to glBlendFuncSeparate):
```rust
...
    pub src_rgb: BlendFactor,
    pub dst_rgb: BlendFactor,
    pub src_alpha: BlendFactor,
    pub dst_alpha: BlendFactor,
```

not-fl3:
Yes, good catch, there should be separate `color_blend` and `alpha_blend`

nokola:
Fixed by above PR

PR #71, opened 2020-04-21
Separate blend for RGB and Alpha and saturate alpha support

nokola:


nokola:
No problem, fixes #55

PR #74, opened 2020-04-27
Support set_blend without resetting pipeline

nokola:
I need to be able to change the blend in my graphics library without re-applying the whole pipeline. e.g. when drawing multiple strokes/paths on top of each other with different blend. 
Something like this:
```rust
fn flush_draw() {
...
    self.ctx.apply_pipeline(&self.pipeline);
    self.ctx.apply_bindings(&self.bindings); 
    for call in calls {
         self.ctx.set_blend(Some(&call.blend_func));
         draw_triangles(call.triangles); // draw some path with blending
    }
... 
}


not-fl3:
Makes sense, thanks!

PR #75, opened 2020-04-28
Update GL_TRUE, GL_FALSE from u32 to u8 to be easier in OpenGL functions

nokola:


not-fl3:
I would try to avoid modifying the bindgen generated code and would prefer to write something like `GL_TRUE as _` or `true as _` in the implementation code instead.

nokola:
Sounds good! I didn't know about "as _" and got tired of typing "as u8" multiple times :)

PR #76, opened 2020-04-28
Add support for GL_UNSIGNED_INT

tversteeg:
It seems this was missing, I want to use it to encode colors in a `u32`.

### DIFF
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -100,6 +100,10 @@ pub enum VertexFormat {
     Short2,
     Short3,
     Short4,
+    Int1,
+    Int2,
+    Int3,
+    Int4,
     Mat4,
 }
 
@@ -118,6 +122,10 @@ impl VertexFormat {
             VertexFormat::Short2 => 2,
             VertexFormat::Short3 => 3,
             VertexFormat::Short4 => 4,
+            VertexFormat::Int1 => 1,
+            VertexFormat::Int2 => 2,
+            VertexFormat::Int3 => 3,
+            VertexFormat::Int4 => 4,
             VertexFormat::Mat4 => 16,
         }
     }
@@ -136,6 +144,10 @@ impl VertexFormat {
             VertexFormat::Short2 => 2 * 2,
             VertexFormat::Short3 => 3 * 2,
             VertexFormat::Short4 => 4 * 2,
+            VertexFormat::Int1 => 1 * 4,
+            VertexFormat::Int2 => 2 * 4,
+            VertexFormat::Int3 => 3 * 4,
+            VertexFormat::Int4 => 4 * 4,
             VertexFormat::Mat4 => 16 * 4,
         }
     }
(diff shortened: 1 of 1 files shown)


nokola:
I think GL_UNSIGNED_INT is WebGL2 only per [this site](https://webgl2fundamentals.org/webgl/lessons/webgl2-whats-new.html).
Do we need any #cfg or capabilities check?

@not-fl3 and I discussed before having a way to check the underlying system capabilities e.g. ctx.supports(EnumValue) to enable support for more platform features while still having support for "lowest common denominator" platforms.

I wonder - how can we make the use of capabilities that are only available on some platforms a compile failure rather than runtime failure? This will help users of miniquad remain confident their code will run on the platforms they choose at compile time.
Perhaps using something like `#[cfg(feature = "webgl2")]`. What is your opinion?

tversteeg:
It seems to be supported or is this the wrong documentation? https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/Types

not-fl3:
@nokola 

My guess - in WebGL2 you can do `attribute int Attr[3];` and then use it as an integer, while in WebGL1 you cant. 

but you still can have integer vertex attributes on the CPU side and receive them in the shader as `attribute vec3 Attr;`

At least it works like this for uniforms.

nokola:
I think it's the wrong documentation. I looked at the documentation from the same site: https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/vertexAttribPointer

> void gl.vertexAttribPointer(index, size, type, normalized, stride, offset);
...
> type
> A GLenum specifying the data type of each component in the array. Possible values:
> * gl.BYTE: signed 8-bit integer, with values in [-128, 127]
> * gl.SHORT: signed 16-bit integer, with values in [-32768, 32767]
> * gl.UNSIGNED_BYTE: unsigned 8-bit integer, with values in [0, 255]
> * gl.UNSIGNED_SHORT: unsigned 16-bit integer, with values in [0, 65535]
> * gl.FLOAT: 32-bit IEEE floating point number
> When using a WebGL 2 context, the following values are available additionally:
> * gl.HALF_FLOAT: 16-bit IEEE floating point number
 
The problem with WebGL2 is that it's not available on iPad and most Android phones.

After more digging, just found the official doc! Adding here for reference: 
https://www.khronos.org/registry/OpenGL/specs/es/2.0/es_full_spec_2.0.pdf#nameddest=section-4.4.3
image

from page 12 of the doc:
GL Type Minimum Description
Bit Width
boolean 1 Boolean
**byte 8 Signed binary integer**
**ubyte 8 Unsigned binary integer**
char 8 characters making up strings
**short 16 Signed 2’s complement binary integer**
**ushort 16 Unsigned binary integer**
int 32 Signed 2’s complement binary integer
uint 32 Unsigned binary integer
**fixed 32 Signed 2’s complement 16.16 scaled**
integer
sizei 32 Non-negative binary integer size
enum 32 Enumerated binary integer value
intptr ptrbits Signed 2’s complement binary integer
sizeiptr ptrbits Non-negative binary integer size
bitfield 32 Bit field
**float 32 Floating-point value**
clampf 32 Floating-point value clamped to [0, 1]

nokola:
> My guess - in WebGL2 you can do attribute int Attr[3]; and then use it as an integer, while in WebGL1 you cant.

Just read your comment after pasting the links to docs - thanks for clarifying @not-fl3!

PR #79, opened 2020-05-02
Stencil test support

nokola:
Contains the pipeline options for stencil. Will implement the actual functionality in the next PR, because it's easier to manage the code.

### DIFF
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -287,7 +287,6 @@ pub struct StencilState {
 
 #[derive(Debug, Copy, Clone, PartialEq)]
 pub struct StencilFaceState {
-    // gl.stencilOpSeparate(face, fail_op, depth_fail_op, pass_op)
     /// Operation to use when stencil test fails
     pub fail_op: StencilOp,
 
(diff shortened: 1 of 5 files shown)


not-fl3:
Nice, I like the `StencilState` struct!

src/graphics.rs
```diff
@@ -279,6 +279,70 @@ pub struct BlendState {
...
+    pub fail_op: StencilOp,
+    
+    /// Operation to use when stencil test passes, but depth test fails
+    pub depth_fail_op: StencilOp,
+    
+    /// Operation to use when both stencil and depth test pass, 
+    /// or when stencil pass and no depth or depth disabled
+    pub pass_op: StencilOp,
+    
+    // gl.stencilFuncSeparate(face, test_func, test_ref, test_mask);
```
not-fl3, line 301: Taking future multiple backends into account it will not always be `gl.*`, the actual implementation will depend on the backend.
So I would try to avoid references to GL in the comments to avoid frustration and leave the room to change the implementation.

not-fl3:
Hm, is it really better to merge this to master? Maybe keep it in the working branch until it's going to be complete feature? I do not mind merging as is though, just wondering why is it better?

nokola:
> Nice, I like the `StencilState` struct!

Thanks! Added all the functions too. Verified `cargo run --example quad` works (on Windows).
Fixing the failure on windows-gnu now

not-fl3:
Oh, ok, I see the functions, no questions than :)

src/graphics.rs
nokola: fixed

not-fl3:
Awesome, great job, thanks!

PR #80, opened 2020-05-02
Fix WASM glStencil...() functions

nokola:
hmm, I forgot to copy the gl.js from my running folder back to sapp-wasm/js/gl.js.

Just re-tested WASM from clean branch and it failed. This is the fix.

Wondering if there's a way to add something else to GitHub actions to ensure samples will fail in case of issues like the one below.

**Edit:** or maybe we treat these type of issues as one-off and I'll be more careful next time.

### DIFF
--- a/native/sapp-wasm/js/gl.js
+++ b/native/sapp-wasm/js/gl.js
@@ -695,13 +695,13 @@ var importObject = {
             GL.shaders[id] = gl.createShader(shaderType);
             return id;
         },
-        stencilFuncSeparate: function (face, func, ref_, mask) {
+        glStencilFuncSeparate: function (face, func, ref_, mask) {
             gl.glStencilFuncSeparate(face, func, ref_, mask);
         },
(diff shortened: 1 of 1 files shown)


nokola:
We could add a one-line cargo_test_wasm() command that will copy gl.js and other files in an output folder, start server and test. Added benefit is that for people who start with miniquad for the first time, will be able to have a one-liner to run WASM.

nokola:
Also sorry for breaking it! Will check with clean branch next time!

not-fl3:
No problems, there are too many platforms and CI is far from perfect :) 

It would be nice to somehow check the wasm on CI, but I am not really sure how. 
My dream CI setup: run X server with software opengl and make screenshots for all the examples on all the platforms - to check that nothing is visually broken. But not sure that this is possible with GH actions.

nokola:
> My dream CI setup: run X server with software opengl and make screenshots for all the examples on all the platforms

nice! :)

PR #81, opened 2020-05-02
Support front face/cull face operations

nokola:
Verified 
```
cargo run --example quad
```
and 
```
cargo build --example quad --target wasm32-unknown-unknown
``` 
then run quad in browser and see quads. With properly copying gl.js to examples folder :)


not-fl3:
Awesome!

PR #83, opened 2020-05-03
fixed shader link errors causing "MALFORMED ERROR MESSAGE"

koalefant:
I have got this when I exceeded limit of varying vectors, and this is reported during shader linking (some Apple Laptop with Intel UHD Graphics 630 got only 15 MAX_VARYING_VECTORS).
I did not test anything except for my project.

To clarify, when using eprintln! instead of panic!, the error was "data provided contains a nul byte".

### DIFF
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -957,9 +957,14 @@ fn load_shader_internal(
                 &mut max_length as *mut _,
                 error_message.as_mut_ptr() as *mut _,
             );
+            // trim trailing zeros
+            while error_message.last().copied() == Some(0) {
+                error_message.pop();
+            }
 
             let error_message = std::string::String::from_utf8_lossy(&error_message);
-            panic!("{}", error_message);
+            eprintln!("Shader link error:\n{}", error_message);
+            panic!("can't link shader");
         }
 
         glUseProgram(program);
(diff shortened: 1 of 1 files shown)


src/graphics.rs
```diff
@@ -957,9 +957,14 @@ fn load_shader_internal(
                 &mut max_length as *mut _,
                 error_message.as_mut_ptr() as *mut _,
             );
+            // trim trailing zeros
+            while error_message.last().copied() == Some(0) {
```
not-fl3, line 961: maybe 
```let error_message = std::string::String::from_utf8_lossy(&error_message[0..max_length as usize]); ``` will do the job of trimming zeroes as well?

koalefant: I think glGetProgramInfoLog does not write max_length in gl.js, but I am not insisting on the implementation - consider this to be an issue ticket :smile:.

koalefant: I also tried using GL_INFO_LOG_LENGTH, but it does not seem to work with a program object

not-fl3: oh, ok, my bad :) 
Going to merge it as is and fix gl.js later

koalefant: thank you!

not-fl3: Done: https://github.com/not-fl3/miniquad/pull/84

PR #87, opened 2020-05-10
Cleanup pipeline update code, add set_cull_face, set_color_write to s…

nokola:
…upport easier drawing for concave polygons

### DIFF
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -359,6 +359,7 @@ struct GlCache {
     blend: Option<BlendState>,
     stencil: Option<StencilState>,
     color_write: ColorMask,
+    cull_face: CullFace,
     attributes: [Option<CachedAttribute>; MAX_VERTEX_ATTRIBUTES],
 }
 
@@ -547,6 +548,7 @@ impl Context {
                     blend: None,
                     stencil: None,
                     color_write: (true, true, true, true),
+                    cull_face: CullFace::Nothing,
                     stored_texture: 0,
                     textures: [0; MAX_SHADERSTAGE_IMAGES],
                     attributes: [None; MAX_VERTEX_ATTRIBUTES],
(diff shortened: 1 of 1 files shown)


src/graphics.rs
```diff
@@ -623,20 +611,40 @@ impl Context {
...
         }
+        self.cache.cull_face = cull_face;
+    }
 
-        let pipeline = &self.pipelines[pipeline.0];
-        if self.cache.color_write != pipeline.params.color_write {
-            let (r, g, b, a) = pipeline.params.color_write;
-            unsafe { glColorMask(r as _, g as _, b as _, a as _) }
-            self.cache.color_write = pipeline.params.color_write;
+    pub fn set_color_write(&mut self, color_write: ColorMask) {
```
not-fl3, line 641: I am a little bit concerned about making those methods `pub`. I am not really against it right now, but I am not sure that this will survive other backends, where pipelines are more real than in opengl. 

Also, they break the idea that render state is completely describable in declarative way as renderpass/pipeline/binding, wich is nice.

nokola: Yes, I have similar concerns too. We should balance between "pub" and ability to set the pipeline easily. I try to avoid the pub as much as possible. 

Not sure what's the best way to do it yet - perhaps when we have Metal support it will be clearer if it's better to change these pub back to private.

I'm not very familiar with how other frameworks handle similar situations (although reading about it more these days.) If you have suggestions, please mention! I also looked at intel's fastuidraw library that has an "uber shader" and advertises small amount of draw calls.

**Do you know of anything that's using sokol and drawing non-convex polygons with holes? I can look to see how they use sokol there.**


In the concave polygon draw code currently I have something like:
```
set_pipeline(...)
... 
draw_polygon:
    update stencil()
    draw()
    update_cull_face()
    draw()
    disable_stencil()
    draw()
```
I'll look into making it simpler/less draw calls.

not-fl3: I made a quick look on metal docs and apparently it is possible to keep those methods public even with metal. 
In sokol/oryol IIRC you will end up with 
```
draw_polygon:
  set_pipeline(stencil_pipeline)
  draw()
  set_piepline(..)
  draw()
```

Same with wgpu, gfx and I guess most vulkan-based APIs. Which is sad for that particular use case, so while we could keep this public - why not?

nokola: Sounds good to keep public.

PR #88, opened 2020-05-10
RenderPass::delete

not-fl3:
Now post_processing example can run without gpu memory leak


tversteeg:
When I run the example of this PR in WASM I get the following warning in the browser console:

`WebGL warning: drawElementsInstanced: Index buffer too small.`

I also don't see the background color or the cube.

not-fl3:
Works on my machine for both wasm and windows(at least x86_64-pc-window-gnu with wine)!

PR #89, opened 2020-05-11
Support different upload format than texture format

nokola:
Can now upload alpha pixel format (1 byte per pixel) to RGBA8 texture.
Breaking changes to Texture class.

Very useful to upload effectively Alpha texture data to platforms that only support RGBA8 (like WebGL1) without having to make 4x larger array (4 bytes per pixel vs 1 byte per pixel) on the CPU side

### DIFF
--- a/examples/post_processing.rs
+++ b/examples/post_processing.rs
@@ -1,0 +1,304 @@
+use miniquad::*;
+
+use glam::{vec3, Mat4};
+
+struct Stage {
+    post_processing_pipeline: Pipeline,
+    post_processing_bind: Bindings,
+    offscreen_pipeline: Pipeline,
+    offscreen_bind: Bindings,
+    offscreen_pass: RenderPass,
+    rx: f32,
+    ry: f32,
+}
+
+impl Stage {
+    pub fn new(ctx: &mut Context) -> Stage {
+        let color_img = Texture::new_render_texture(
+            ctx,
+            TextureParams {
+                width: 256,
+                height: 256,
+                format: TextureFormat::RGBA8,
+                ..Default::default()
+            },
+        );
+        let depth_img = Texture::new_render_texture(
+            ctx,
+            TextureParams {
+                width: 256,
+                height: 256,
+                format: TextureFormat::Depth,
+                ..Default::default()
+            },
+        );
+
+        let offscreen_pass = RenderPass::new(ctx, color_img, depth_img);
+
+        #[rustfmt::skip]
+        let vertices: &[f32] = &[
+            /* pos               color                   uvs */
+            -1.0, -1.0, -1.0,    1.0, 0.5, 0.5, 1.0,     0.0, 0.0,
+             1.0, -1.0, -1.0,    1.0, 0.5, 0.5, 1.0,     1.0, 0.0,
+             1.0,  1.0, -1.0,    1.0, 0.5, 0.5, 1.0,     1.0, 1.0,
+            -1.0,  1.0, -1.0,    1.0, 0.5, 0.5, 1.0,     0.0, 1.0,
+    
+            -1.0, -1.0,  1.0,    0.5, 1.0, 0.5, 1.0,     0.0, 0.0, 
+             1.0, -1.0,  1.0,    0.5, 1.0, 0.5, 1.0,     1.0, 0.0,
+             1.0,  1.0,  1.0,    0.5, 1.0, 0.5, 1.0,     1.0, 1.0,
+            -1.0,  1.0,  1.0,    0.5, 1.0, 0.5, 1.0,     0.0, 1.0,
+    
+            -1.0, -1.0, -1.0,    0.5, 0.5, 1.0, 1.0,     0.0, 0.0,
+            -1.0,  1.0, -1.0,    0.5, 0.5, 1.0, 1.0,     1.0, 0.0,
+            -1.0,  1.0,  1.0,    0.5, 0.5, 1.0, 1.0,     1.0, 1.0,
+            -1.0, -1.0,  1.0,    0.5, 0.5, 1.0, 1.0,     0.0, 1.0,
+    
+             1.0, -1.0, -1.0,    1.0, 0.5, 0.0, 1.0,     0.0, 0.0,
+             1.0,  1.0, -1.0,    1.0, 0.5, 0.0, 1.0,     1.0, 0.0,
+             1.0,  1.0,  1.0,    1.0, 0.5, 0.0, 1.0,     1.0, 1.0,
+             1.0, -1.0,  1.0,    1.0, 0.5, 0.0, 1.0,     0.0, 1.0,
+    
+            -1.0, -1.0, -1.0,    0.0, 0.5, 1.0, 1.0,     0.0, 0.0,
+            -1.0, -1.0,  1.0,    0.0, 0.5, 1.0, 1.0,     1.0, 0.0,
+             1.0, -1.0,  1.0,    0.0, 0.5, 1.0, 1.0,     1.0, 1.0,
+             1.0, -1.0, -1.0,    0.0, 0.5, 1.0, 1.0,     0.0, 1.0,
+    
+            -1.0,  1.0, -1.0,    1.0, 0.0, 0.5, 1.0,     0.0, 0.0,
+            -1.0,  1.0,  1.0,    1.0, 0.0, 0.5, 1.0,     1.0, 0.0,
+             1.0,  1.0,  1.0,    1.0, 0.0, 0.5, 1.0,     1.0, 1.0,
+             1.0,  1.0, -1.0,    1.0, 0.0, 0.5, 1.0,     0.0, 1.0
+        ];
+
+        let vertex_buffer = Buffer::immutable(ctx, BufferType::VertexBuffer, &vertices);
+
+        #[rustfmt::skip]
+        let indices: &[u16] = &[
+            0, 1, 2,  0, 2, 3,
+            6, 5, 4,  7, 6, 4,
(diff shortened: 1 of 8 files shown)


nokola:
Open to suggestions for improvements (as always!)

nokola:
Will wait to fix the build failures, would like to discuss first with @not-fl3 if this change is the right approach:
1. Would there be issue with platform support? (e.g. raspberry Pi? I know webgl1 is OK)
2. Should we do breaking change in the Texture:: functions or add new ones? I personally think breaking changes to the existing APIs are better because the new Texture::update..() functions document the miniquad capabilities better if they have extra pixel_format parameter

not-fl3:
Hmm, those structs look so similar that it really confuses me wich one is wich.
I do not have any ideas on how to avoid this, will think about this tomorrow!

nokola:
> Hmm, those structs look so similar that it really confuses me wich one is wich.
> I do not have any ideas on how to avoid this, will think about this tomorrow!

Thanks for the reply! 
Alternative idea: Keep one format struct that has ::Alpha but `panic!` when trying to create Alpha texture until we have at some point ability to detect capabilities.

not-fl3:
>  Alternative idea: Keep one format struct that has ::Alpha but panic! when trying to create Alpha texture until we have at some point ability to detect capabilities.

I would definetly prefer this!

nokola:
done! Yes, I can see it feels less confusing.

nokola:
PR is ready from my side (open to modifications, of course.) Verified WASM and Windows MSVC work.

not-fl3:
The problem here - in WebGL1 the only supported single channel texture ins GL_ALPHA, and that single channel is an alpha channel, which will require additional set of shaders :( 

By myself, I would prefer small runtime and memory overhead and actually copy channels before uploading than writing to sets of shaders :( 
However, I am fine merging this - but with some panic/warning on runtime and clear descrition in comments that this is supported only on WebGL2 (not supported on ios/macos)

not-fl3:
And sorry for a late response when the whole PR is basically done! The PR itself is great and I would like to have this code available for desktop-only users somehow.

nokola:
> The problem here - in WebGL1 the only supported single channel texture ins GL_ALPHA, and that single channel is an alpha channel, which will require additional set of shaders :(

Yes - this is a problem in general, but not in this PR. 
This PR still does not support Alpha textures (see panic! note further below). The only support is uploading Alpha data -> RGBA texture which works on WebGL1 according to docs.
That means a user of miniquad will still have to write single shader for all platforms, because the shader texture is still RGBA even if the data from the CPU is in [u8] alpha format.

Unless I'm mistaken, I think the Alpha data will get copied to each of the R, G, B, A channels on upload to the GPU.

> However, I am fine merging this - but with some panic/warning on runtime

panic is already there - code will panic on Alpha texture creation on all platforms (webgl, desktop...) with "not yet supported because of webgl1" message
These are the panics on new texture creation:
```
assert!(params.format != TextureFormat::Alpha, "TextureFormat::Alpha textures are not 
supported yet. Use TextureFormat::RGBA textures instead and upload TextureFormat::Alpha data");

assert!(params.format == pixel_format, "Different GPU and CPU pixel formats are not supported
when creating texture (OK when updating texture data), because of 
limitations in WebGL1 and not yet implemented support.");
```

A person will still have to write the exact same code for WebGL/desktop/etc. after this PR, until we figure out a good way to query capabilities and support at compile and/or runtime

>And sorry for a late response when the whole PR is basically done! 

No problem, it was good to have time to think about it! I appreciate all the time you put in responding/updating things promptly.

nokola:
If I missed some panic!() etc case, please let me know, thanks!

not-fl3:
Oh, ok, I see, my mistake. Yes, GL_RED is used in From<TextureFormat> but will actually never used because of the panic on ::Alpha. 

In that case - maybe just add something like `pixel_store_aligment: Option<u32>` into `TextureParams`? 
The only usage of PixelFormat::Alpha - is to pass that one param to GL, better to keep that intention clear I think

nokola:
Hmm, will think some more about it.

>In that case - maybe just add something like pixel_store_aligment: Option<u32> into TextureParams?
The only usage of PixelFormat::Alpha - is to pass that one param to GL, better to keep that intention clear I think

Not sure I inderstand the idea:
1. `pixel_store_alignment` is actually used as "pixel read alignment" (how the input data is stored) just for the upload of data to GPU. In my mind `TextureParams` contains data about "texture format/settings on GPU". One can upload multiple types of data (Alpha, RGBA) to the same RGBA texture with different calls to `update(...)`. However I like the simplification of having some extra value in `TextureParams`

> only usage of PixelFormat::Alpha - is to pass that one param to GL

2. `GL_RED` is still used in `update()` (not used in `new()` because of the panic!), also the data size is checked based on whether TextureFormat::Alpha is passed or not in `update()`

Any ideas/suggestions please let me know. I'll take a look at the code again later today.

nokola:
Just to clarify, my goal is:
1. Keep compatibility with all platforms (webgl/etc.) with single source/shader code. Enforce no special code for webgl1 or desktop through panic! until we have explicit way to check capabilities
2. Forward compatibility  - once we add compile and runtime checks for platform, ideally API does not break

not-fl3:
If we will have a field with texture alignment/packing in texture params - I think we will be able to achieve the same results without additional PixelFormat concept.

not-fl3:
That way we will TextureFormat::Alpha at all, but still will be able to save some memory on uploads of single channel textures

nokola:
> If we will have a field with texture alignment/packing in texture params - I think we will be able to achieve the same results without additional PixelFormat concept.

we can do it, however does that mean if pixel_store_alignment == 1, we always assume Alpha format for the input pixels array? Technically, pixel_store_alignment of 1 can be used for uploading Alpha -> RGBA or RGBA -> RGBA as well. If we use pixel_store_alignment of 1 to mean "Alpha" and 4 to mean "RGBA" we're basically defining an enum again, but without having a Rust type for it.
Not sure - maybe I misunderstand your intention, please let me know.

not-fl3:
I am seeing this not as a way to get alpha textures, but as a way to effiecently upload single-channel data into 4-channel gpu texture. From this point of view additional memory configuration makes total sense.

nokola:
Yes, I see the functionality also as a way to efficiently upload single-channel data to 4-channel texture.
What I'm not sure about is how pixel_store_alignment defines how many channels the input data has? IMO, pixel alignment is completely separate from number of channels, because one could have 
Perhaps we should use two variables instead of just pixel_store_alignment: `pixel_size` (1/2/4 bytes) and `pixel_row_alignment` (1/2/4)

Or we could use one variable: `pixel_size` and assume:
if pixel_size == 1, alignment is 1
if pixel_size == 2 or 4, alignment is 4

I personally prefer having two variables instead of one to make state more obvious.

not-fl3:
I am totally fine with any amount of variables in TextureParams! My point - let's make them as explicit as possible. Maybe add some field like Option<TextureMemoryLayout> with two fields? 
But I actually do not have any opinion here, just throwing out some ideas

nokola:
I'm planning to abandon this PR, until we figure out WebGL1/etc differences. Will create another pr to add support for just the GL `glPixelStorei` functions since for sure we'll need it in the (maybe near?) future.

**Until the full functionality is there, adding some thoughts here to help with the future design:**

for reference, here's the update() function before changes:
```rust
pub fn update(&self, ctx: &mut Context, bytes: &[u8]) {
```
WebGL function looks like:
```
void gl.texImage2D(target, level, internalformat, width, height, border, format, type, ArrayBufferView? pixels);
```
and OpenGL looks like:
```c
void glTexImage2D(GLenum target,
 	GLint level,
 	GLint internalformat,
 	GLsizei width,
 	GLsizei height,
 	GLint border,
 	GLenum format, // This is part of the format I call CPUDataFormat below
 	GLenum type, // This is part of the format I call CPUDataFormat below
 	const void * data);
```

**Proposal**
1. The CPU texture data format shouldn't be stored in TextureParams, since it's used only during calls to update() and the CPU data format is tied to `pixels`, not tied to the texture during update().

2. Following from (1.) above: either change the existing `update` and `new` calls or add new ones, something like this:
```rust
pub fn update(&self, ctx: &mut Context, bytes: &[u8], bytes_format: Option<CPUDataFormat> ) {
```
where the bytes_format types look something like this:

```rust
enum CPUPixelFormat {
    Red, // not supported on WebGL1
    Alpha,
    RG, // not supported on WebGL1
    RGB,
    RGBA,
...
}

enum CPUDataType {
    UByte,
    Byte,
    UShort,
    UShort_5_6_5,
    ...
}

/// Contains information about format of texture data during upload to GPU texture or download from GPU texture
/// Used to enable uploading of CPU textures with different channels per pixel to GPU
/// # Example
/// bytes_per_pixel == 1 and row_start_alignment == 1 specifies a single-channel texture
/// that may be uploaded to a 4-channel RGBA texture.
/// In this example, the single channel will be copied in the R, G, B, and A components of the texture.
#[derive(Debug, Copy, Clone)]
pub struct CPUDataFormat {
    /// Format of the channels stored in the array passed to GPU
    pub pixel_format: CPUPixelFormat,
    /// Data type stored in each channel
    pub data_type: CPUDataType,
    /// Specifies the alignment for the start of each pixel row in memory.
    /// The allowable values are 1 (byte-alignment), 2, 4 (default), and 8
    pub row_start_alignment: u32,
}
```

Comments and ideas welcome! I think until we have capabilities detection support at compile time and at runtime I'll have to use the 4x size RGBA arrays instead of single-component arrays.
Will come to discuss on discord too in a bit.

PR #94, opened 2020-05-15
Implement glDeleteFramebuffers for WASM

tversteeg:
I'm not sure how to test this function and if it's properly implemented, so please review it carefully.

It was thrown as an error in the console that this function was missing.

### DIFF
--- a/native/sapp-wasm/js/gl.js
+++ b/native/sapp-wasm/js/gl.js
@@ -811,6 +811,20 @@ var importObject = {
                 GL.buffers[id] = null;
             }
         },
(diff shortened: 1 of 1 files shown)


native/sapp-wasm/js/gl.js
```diff
@@ -811,6 +811,20 @@ var importObject = {
                 GL.buffers[id] = null;
             }
         },
+        glDeleteFramebuffers: function (n, buffers) {
```
not-fl3, line 814: Looks close enough to `library_webgl.js` original version :)

not-fl3:
Thanks, LGTM! :)

PR #96, opened 2020-05-16
Rename WASM time to now

tversteeg:
This allows my PR to the instant library to work out of the box without any manual patches to the JS files: https://github.com/sebcrozet/instant/pull/14

It's also more consistent with the API itself since the Rust calling function is also `now()`.

As an aside the following can be added to your `Cargo.toml` to work with libraries depending on `instant` before the PR is merged:

```toml
[patch.crates-io]
instant = { git = "https://github.com/tversteeg/instant.git" }
```


not-fl3:
Nice, thanks!

PR #99, opened 2020-05-17
miniquad: added mapping for most of the SAPP-keys

koalefant:
these are still missing:
APOSTROPHE: return 39;
SLASH: return 47;
GRAVE_ACCENT: return 96;
WORLD_1: return 161;
WORLD_2: return 162;

I am also keeping F11, F12 not mapped as I got complains from people who are using F11, and use F12 myself for debugging


not-fl3:
Nice, thanks!

PR #100, opened 2020-05-17
Support row alignment for texture upload/download

nokola:


nokola:
missed important size() calculation - marked the PR as draft. total array size now depends on (row_alignment, width, height), not just (width, height)

not-fl3:
LGTM! 
Thanks!

PR #104, opened 2020-05-20
Rename WASM time to now

tversteeg:
I think something went wrong with merging because these changes were gone.


not-fl3:
Yes, I do remember those changes. This is weird!

Issue #54, opened 2020-03-31
Support error handling in Shader::new() API


nokola:
Today miniquad will panic if invalid shader code is specified in Shader::new()

It may be useful for callers to gracefully show error message instead of panic!() especially for tools that get arbitrary shader text (e.g. from the user, like https://shadertoy.com)

I propose to add Result<..> to Shader::new() to support error handling

nokola:
Also note: I'm opening all these issues since I'm going through the nvg-miniquad port and adding tracking issues for whatever the difference is between the nvg expected APIs and miniquad, if the new APIs will improve miniquad.

not-fl3:
I agree, `Shader::new()` should return Result with error message.

nokola:
Edit: renamed title to be more generic since I found panic-s in several places:
```rust
Shader::new()
get_uniform_location()
apply_bindings()
Pipeline::with_params()
```

Would be nice to have error propagation in general for miniquad. Perhaps that means `EventHandler` updates to have the functions there return Result as well.

PR #108, opened 2020-05-24
graphics: Return Result instead of panic for shaders compilation

not-fl3:
This breaks backward compatibility - old `Shader::new()`  should be replaced to `Shader::new().unwrap()` for the same behaviour.

But allows to build cool interactive playgrounds!
https://not-fl3.github.io/miniquad-samples/shadertoy.html


nokola:
Also fixes part of #54!
Better to break back compat early than late and error handling is important! Glad this is happening!

PR #114, opened 2020-06-04
sapp_x11: Support "Conf::fullscreen"

not-fl3:
Now "window_free" example runs in full screen! 

Tested with i3 and openbox, looks like it works!


pum-purum-pum-pum:
It works
xorg-server-hwe-18.04 2:1.20.5+git20191008-0ubuntu1~18.04.1

PR #116, opened 2020-06-04
Set repeat flag for key down event

### LINKED ISSUE #22: key_down_event `repeat` flag isn't set on repeat keys
The `repeat` flag doesn't appear to be set, even when a key is being held down and the event is a "repeat" event:

```
    fn key_down_event(
        &mut self,
        _ctx: &mut Context,
        keycode: KeyCode,
        _keymod: KeyMods,
        repeat: bool,
    )
```

pavlukivan:
Closes #22. Can be backported to 0.2


not-fl3:
Thanks!

not-fl3:
published as "0.3.0-alpha.9"

PR #122, opened 2020-06-19
sapp-wasm: call preventDefault() on wheel event to prevent scroll of the longer page when wheel is used

koalefant:


not-fl3:
Thanks!

PR #124, opened 2020-06-28
Disable browser right-click context menu.

Slyklaw:


examples/index.html
```diff
@@ -21,7 +21,7 @@
 </head>
 
 <body>
-    <canvas id="glcanvas" tabindex='1'></canvas>
+    <canvas id="glcanvas" tabindex='1' oncontextmenu="return false;></canvas>
```
lukabavdaz, line 24: Forgot a quotation mark after `false;`

Slyklaw: PR updated.

PR #128, opened 2020-07-19
Separate JS WASM files into three modules: gl.js, sapp.js and wasm.js

rellfy:
This PR organises the JS components for WASM a bit. It creates a simple `Wasm` class in `wasm.js` which is accessible through the current page via `window.wasm` and loads functions into `importObject.env` via `window.wasmImports`, so multiple scripts can add import functions - which is used in both `gl.js` and `sapp.js`.

The example `index.html` accounts for those changes by adding the two new scripts.

The `Wasm` class could be further simplified by removing the `plugins` methods and replacing the `plugin.on_init()` calls with events, but this could break things if anything currently depends on `plugins` so I didn't want to do that now.

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -1,6 +1,4 @@
-<!doctype html>
-<html lang="en">
-
+<html>
 <head>
     <meta charset="utf-8">
     <title>TITLE</title>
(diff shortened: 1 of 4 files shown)


koalefant:
What is the motivation behind this change?
Each additional module is an extra fetch and extra latency for application loading.

rellfy:
> What is the motivation behind this change?
> Each additional module is an extra fetch and extra latency for application loading.

For purposes of clean code, as `gl.js` is a mess currently. My commits don't solve the issue completely, but it's possibly a start.
About the loading times: this is the source. It can be distributed as a single bundled and minified file as it is currently, but that doesn't mean the source needs to be a single huge file.

not-fl3:
A lot of work done here, appreciate this! 

However I can really see the benefits of this change but can see some problems - now we will need to store 3 files instead of 1 and keep versions in sync of 3 files. 

gl.js deserve some refactoring for sure, but I see this PR: https://github.com/not-fl3/miniquad/pull/97 as a first step of that refactoring.

PR #131, opened 2020-07-31
Fix setting 3f uniforms

mbirtwell:
Just a typo in the array construction I think.

Previously produced errors like:

   gl.js:643 WebGL: INVALID_VALUE: uniform3fv: invalid size
        glUniform3fv @ gl.js:643


not-fl3:
Thanks!

PR #133, opened 2020-08-12
Remove the dependency on libc

Gordon-F:
Replace libc with cty for linux and dummy.


Gordon-F:
@not-fl3 Can you review this pr?

not-fl3:
I am not familiar enough with `cty` to make a decision. Is it really easy to tell bindgen to use cty instead of libc? Is it really better than libc?

Gordon-F:
@not-fl3 [cty](https://github.com/japaric/cty) just provide type aliases to C types. Currently we use libc just for types. This changes also improve compile time.

PR #134, opened 2020-08-18
Added graphics::ElapsedQuery for GPU-profiling

koalefant:
It was only tested on Linux and WASM and problably won't compile yet on other platforms. 

Is it something that we would like to have?


not-fl3:
Yes, absolutely. 

I could help with compilation errors or just fix them later

koalefant:
I will get Windows fixed at a later point, and maybe Android, but I don't have an Apple workstation at the moment.

koalefant:
I overlooked one important detail: EXT_disjoint_timer_query does not work on Firefox. So for this to be useful one would need ElapsedTimer::is_supported() :disappointed:

koalefant:
Added Windows support.

native/sapp-windows/src/gl.rs
```diff
@@ -598,5 +602,11 @@ gl_loader!(
     fn glGenVertexArrays(n: GLsizei, arrays: *mut GLuint) -> (),
     fn glFrontFace(mode: GLenum) -> (),
     fn glCullFace(mode: GLenum) -> (),
-    fn glGenTextures(n: GLsizei, textures: *mut GLuint) -> ()
+    fn glGenTextures(n: GLsizei, textures: *mut GLuint) -> (),
```
not-fl3, line 605: The new gl generator is so much better compared to the nonsense we had before, isn't it? :D

koalefant: Yes, it was a pleasure to do this change with the new loader.

not-fl3:
> I overlooked one important detail: EXT_disjoint_timer_query does not work on Firefox. So for this to be useful one would need ElapsedTimer::is_supported() disappointed

Found it just when I started profiling weird firefox only FPS drop :D  

But fortunately is not actually completely broken in firefox: with "webgl.enable-privileged-extensions:true" EXT_disjoint_timer_query still works!

koalefant:
> webgl.enable-privileged-extensions:true
Yay, now I can profile on Firefox as well! 😊 Did not know about this option.

PR #135, opened 2020-08-28
Metal backend

Gordon-F:
Draft Metal backend implementation.

Current goals:
- [ ] Implement MacOS Metal backend without public api changes (or with minimal changes)
- [ ] Implement minimal features set (without compute shaders, multithreading and etc.)
- [ ] Use native Metal shaders (Currently we can use [CLI spirv-cross](https://github.com/KhronosGroup/SPIRV-Cross). But also we can use [naga](https://github.com/gfx-rs/naga))


.gitignore
```diff
@@ -1,3 +1,5 @@
 /target
 **/*.rs.bk
-Cargo.lock
\ No newline at end of file
+Cargo.lock
+*.vscode
```
akxcv, line 4: AFAIK, files like `.vscode` and `.idea` are usually put in one's global gitignore.

.gitignore
```diff
@@ -1,3 +1,5 @@
 /target
 **/*.rs.bk
-Cargo.lock
\ No newline at end of file
+Cargo.lock
```
akxcv, line 3: Should we really ignore `Cargo.lock`?

.gitignore
not-fl3: Yes, I agree, user's global gitignore is a good place for user-specific IDE stuff

.gitignore
not-fl3: I am not quite sure about this, usually the good rule is - ignore Cargo.lock for libs, version control Cargo.lock for binaries

Gordon-F:
Ok, i think it's enough for first small step. At least it can run quad example on Mac, iOS and iOS Simulator 😄

not-fl3:
Hi from 2023! haha

Took me a little bit, but I finally merged this very PR: https://github.com/not-fl3/miniquad/pull/338
Changed metal-rs to objc, and reimplemented a few things here and there, but, basically, new miniquad's metal backend is based on this very PR! 

Thank you! 

:+1:

PR #138, opened 2020-09-23
sapp-windows: missing null termination for CreateWindowExW arguments

koalefant:
Otherwise CreateWindowExW fails unpredictably.


not-fl3:
Ohh that was bad, thanks a lot!

PR #142, opened 2020-10-01
graphics: Remove all 'static references from ShaderMeta

not-fl3:
**What happened:**

```
-    pub const META: ShaderMeta = ShaderMeta {
-        images: &["tex"],
-        uniforms: UniformBlockLayout {
-            uniforms: &[UniformDesc::new("resolution", UniformType::Float2)],
-        },
-    };
+    pub fn meta() -> ShaderMeta {
+        ShaderMeta {
+            images: vec!["tex".to_string()],
+            uniforms: UniformBlockLayout {
+                uniforms: vec![UniformDesc::new("resolution", UniformType::Float2)],
+            },
+        }
+    }

```

**Why:** 

&'static arrays were really unfriendly for runtime shader meta building.

**Problems:** 

- got couple of new allocations for arrays
- awkward looking `.to_string()` calls
- it is very very breaking change

**Alternative:** 

Keep ShaderMeta as is, but change Shader signature to

```
    pub fn new(
        ctx: &mut Context,
        vertex_shader: &str,
        fragment_shader: &str,
        meta: impl into<ShaderMetaRuntime>,
    ) -> Result<Shader, ShaderError> {
```
and introduce `ShaderMetaRuntime` struct usable in runtime. 
This will resolve all the issues but will introduce a bunch of new structs and probably will introduce some confusion on two sets on very same structs.

### DIFF
--- a/examples/blobs.rs
+++ b/examples/blobs.rs
@@ -40,7 +40,7 @@ impl Stage {
             images: vec![],
         };
 
-        let shader = Shader::new(ctx, shader::VERTEX, shader::FRAGMENT, shader::META).unwrap();
+        let shader = Shader::new(ctx, shader::VERTEX, shader::FRAGMENT, shader::meta()).unwrap();
 
         let pipeline = Pipeline::new(
             ctx,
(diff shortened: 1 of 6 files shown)


not-fl3:
Let's say that alpha is an alpha for a reason and merge it. 
Maybe will roll back later though.
 
@koalefant I guess it will affect you the most, any arguments against this?

koalefant:
@not-fl3 looks good to me :+1:

PR #154, opened 2020-11-07
Add set_window_size api for Windows

Bombfuse:
Quick API for setting window size for Windows OS.

I plan on adding implementation for this to Mac, Linux, and WASM as well.

Wasm will just call out to JS to resize the canvas.


examples/window_manipulation.rs
```diff
@@ -0,0 +1,35 @@
...
+            } else if keycode == KeyCode::E {
+                ctx.set_window_size(800, 500);
+            } else if keycode == KeyCode::R {
+                ctx.set_window_size(1600, 900);
+            }
+        }
+}
+
+fn main() {
+    miniquad::start(conf::Conf::default(), |ctx| UserData::owning(Stage { is_fullscreen: false, }, ctx));
```
not-fl3, line 34: I guess you want to run rustfmt here :)

native/sapp-wasm/js/gl.js
```diff
@@ -1170,6 +1170,18 @@ var importObject = {
...
+            if (!fullscreen) {
+                document.exitFullscreen();
+            } else {
+                canvas.requestFullscreen();
+            }
+        },
+        sapp_set_window_size: function(new_width, new_height) {
+            canvas.width = new_width;
+            canvas.height = new_height;
+            resize(canvas, wasm_exports.resize);
```
not-fl3, line 1184: does it really works? If I remember correctly resize will be called on the window resize, so it looks like set_window_size will work only until the next window resize JS event?

native/sapp-windows/src/lib.rs
```diff
@@ -1023,21 +1087,32 @@ unsafe fn create_window() {
...
         bottom: 0,
     };
+
     if _sapp.desc.fullscreen {
         win_style = WS_POPUP | WS_SYSMENU | WS_VISIBLE;
         rect.right = GetSystemMetrics(SM_CXSCREEN);
         rect.bottom = GetSystemMetrics(SM_CYSCREEN);
     } else {
-        win_style = WS_CLIPSIBLINGS
+        win_style = if _sapp.desc.window_resizable {
```
not-fl3, line 1096: Maybe do keep the common window flags outside of if? 

Something like this: 
win_style = WS_CLIPSIBLINGS | WS_CLIPCHILDREN | WS_CAPTION ..;
if _sapp.desc.window_resizable {
  win_style |= WS_SIZEBOX
}

examples/window_manipulation.rs
not-fl3: And I would actually avoid showcasing this API until it works on linux/mac as well - it would be a little frustrating to see the example that doesnt work :)

native/sapp-wasm/js/gl.js
```diff
@@ -1170,6 +1170,18 @@ var importObject = {
             } else {
                 document.exitPointerLock();
             }
+        },
+        sapp_set_fullscreen: function(fullscreen) {
+            if (!fullscreen) {
+                document.exitFullscreen();
+            } else {
+                canvas.requestFullscreen();
```
not-fl3, line 1178: I did not really tried it myself so it is just a question - does the browser send resize event after going fullscreen? From quick googling everyone do some things with canvas size after requestFullscreen, but if it works without - good for us!

native/sapp-wasm/js/gl.js
Bombfuse: Yah this works, I tested it out using the `index.html` in the repo and by removing `width: 100%, height: 100%`.

native/sapp-windows/src/lib.rs
Bombfuse: good callout! I can update this

native/sapp-wasm/js/gl.js
Bombfuse: I didn't notice any weird behavior, it seems like it just works. One thing to note though, is that if you go fullscreen in windows, and then resize the window, you get the classic `borderless` affect that is in a lot of games.

Bombfuse:
After discussion on discord, the plan is to

* Leave `set_fullscreen` as it is
* Disable `set_window_size` during fullscreen mode
* Warn that `set_window_size` does not work during fullscreen
* Update `set_window_size` to update size in the same window position
* Open a ticket tracking that `set_window_size` updates video resolution mode when called during fullscreen

Bombfuse:
Replacing this PR with https://github.com/not-fl3/miniquad/pull/189

PR #159, opened 2020-12-18
Use strict mode in js

maciejhirsz:
[Strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode) is the de-facto standard JS these days, and most if not all JS bundlers / transpilers enforce it.

There are performance advantages, and it helps avoid accidentally overwriting things in global scope (you can still set things in global scope by using `window.GLOBAL = ...` instead of just `GLOBAL = ...`).


not-fl3:
Thank you!

PR #160, opened 2020-12-20
Fix wasm32 dependencies in Cargo.toml

vnetserg:
It looks like this patch is the only thing needed to make microquad compile for `wasm32-wasi` target.

P.S. Great work with microquad and macroquad, I love it!


not-fl3:
wow, thats a good news, never tried wasm32-wasi :)

Thanks for PR!

PR #161, opened 2020-12-23
sapp-wasm: Fix touch event listeners in js

Drewol:
Tested touch on my phone and in chromium's emulated phone environment and it did not seem to work. The JS event handlers were throwing errors, this should fix that.


not-fl3:
Thanks!

PR #162, opened 2020-12-26
Random doc comments

maciejhirsz:
Nothing fancy, and by no means complete, but better than nothing :)

I'm not sure how much detail needs to be given on inner workings of miniquad as an abstraction layer, on one hand it would be useful to describe how everything works, on the other hand a lot of that is rendundant with OpenGL, so maybe linking to relevant OpenGL docs is the way to go?

### DIFF
--- a/src/conf.rs
+++ b/src/conf.rs
@@ -65,16 +65,27 @@ pub enum Loading {
 pub struct Conf {
     pub cache: Cache,
     pub loading: Loading,
+    /// Title of the window, defaults to an empty string.
     pub window_title: String,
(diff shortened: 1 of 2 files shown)


not-fl3:
Thank you, docs are super important and lacking! 

> maybe linking to relevant OpenGL docs is the way to go

Taking soon to be merged Metal PR into account,  linking to OpenGL docs may not be really helpfull. And lot of things are abstracted away and do not exactly maps to OpenGL. So I would try to avoid using to much OpenGL specific things in the docs.

PR #164, opened 2020-12-31
Cursor grab on Windows

cedric-h:


native/sapp-windows/src/lib.rs
```diff
@@ -671,11 +737,43 @@ unsafe extern "system" fn win32_wndproc(
...
+                    &mut data as *mut _ as _,
+                    &mut size as *mut _ as _,
+                    std::mem::size_of::<RAWINPUTHEADER>() as _,
+                );
+                if get_succeed as i32 == -1 {
+                    panic!("failed to retrieve raw input data");
+                }
+
+                if data.data.mouse().usFlags & MOUSE_MOVE_ABSOLUTE == 1 {
+                    panic!("grr absolute coords");
```
not-fl3, line 766: shouldnt we just ignore the ABSOLUTE part of the event if it happens to be absolute? (I am not sure how it works, so maybe panic is actually the right thing to do)

native/sapp-windows/src/lib.rs
```diff
@@ -464,7 +468,37 @@ pub unsafe fn sapp_mouse_shown() -> bool {
...
-pub unsafe fn sapp_set_cursor_grab(mut _grab: bool) {}
+pub unsafe fn sapp_set_cursor_grab(grab: bool) {
+    if grab == _sapp.cursor_grabbed {
+        return;
+    }
+
+    _sapp.cursor_grabbed = grab;
+
+    let mut rid: RAWINPUTDEVICE = RAWINPUTDEVICE {
+        usUsagePage: 0x01,
```
not-fl3, line 479: Maybe use constants from winapi here?
 
 winapi::shared::hidusage::HID_USAGE_GENERIC_POINTER and HID_USAGE_GENERIC_MOUSE
 ```
HID_USAGE_GENERIC_POINTER: USAGE = 0x01;
HID_USAGE_GENERIC_MOUSE: USAGE = 0x02;
```

native/sapp-windows/src/lib.rs
cedric-h: yeah, I'm not sure. glfw has a bunch of code to track the last known position and still provide deltas even if the hardware uses absolute coordinates, but I heard hardware very rarely if ever gives off these events. sure enough, I tested all of the mice I had and they all gave off relative mouse coords. I think support for absolute coords should only be added if someone is actually running into this panic, because I think ABSOLUTE is just legacy at this point.

not-fl3: Got it, sounds reasonable!

PR #168, opened 2021-01-21
graphics: Fix texure wrapping

scurest:
Hi. I noticed the TextureParams.wrap field wasn't wired up to the actual GL wrap mode, which was always Clamp. This fixes it. I just mimicked how TextureParams.filter works.

### DIFF
--- a/src/graphics/texture.rs
+++ b/src/graphics/texture.rs
@@ -97,8 +97,6 @@ pub enum TextureWrap {
     Mirror,
     /// Samples at coord x + 1 map to coord 1.
     Clamp,
-    /// Same as Mirror, but only for one repetition.
-    MirrorClamp,
 }
 
 #[derive(Clone, Copy, Debug, PartialEq)]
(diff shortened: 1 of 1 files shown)


scurest:
So pretty much only the linux backend has `GL_MIRROR_CLAMP_TO_EDGE`. Should I just add it to the other backends? Also it doesn't seem as widely supported as the others (it only became core in GL 4.4). Does it need some fallback logic?

not-fl3:
Thanks for PR, good catch!

>  GL_MIRROR_CLAMP_TO_EDGE is available only if the GL version is 4.4 or greater. 

yeah, it looks like MIRROR_CLAMP_TO_EDGE is available only on the newest GLs.
I would rather remove `MirrorClamp` from the enum to guarantee that all the enum variants work on all the miniquad-supported platforms.

scurest:
Sounds good, that made everything green.

not-fl3:
Awesome, thanks for PR!

PR #170, opened 2021-01-22
native/sapp-darwin: Fix high-dpi bug

not-fl3:


emilk (approved):
Yeah, this works! Thanks for the super-quick fix!

not-fl3:
Awesome, glad it worked! 

Phew, dodged a bullet here, mac+higdpi are usualy not the best combo :D

Issue #171, opened 2021-01-23
[feature] Set cursor icon


emilk:
For [the egui integration](https://github.com/not-fl3/emigui-miniquad) it would be wonderful if `miniquad` support setting the cursor icon. The icons (currently) required by egui are

``` rust
#[derive(Clone, Copy)]
pub enum CursorIcon {
    Default,
    /// Pointing hand, used for e.g. web links
    PointingHand,
    ResizeHorizontal,
    ResizeNeSw,
    ResizeNwSe,
    ResizeVertical,
    Text,
    /// Used when moving
    Grab,
    Grabbing,
}
```

I realize this is a big ask, but I think changing the mouse cursor when hovering different UI elements is quite important for UX.

This can probably wait until the egui-miniquad integration has grown a bit in popularity (as I hope it will!)

not-fl3:
Now, this is implemented on windows/wasm/linux! https://github.com/not-fl3/miniquad/pull/173

Thanks again, @scurest!

Marking the issue as "Help wanted" as a call for help for mac os implementation :)

scurest:
Also note that I didn't do the grab/grabbing icons OP requested. The big blocker for that is Windows; like I said in #173, I believe those would require embedding our own cursor images to use (for example, I think [this](https://github.com/chromium/chromium/blob/d8b3f8592c5f20cc7a32cca06f3acda19463251d/ui/resources/cursors/hand_grab.cur) is what Chromium uses).

PR #173, opened 2021-02-01
set_mouse_cursor

scurest:
Hey. This is a sketch of an implementation of #171, setting the cursor icon. ~~I'm hoping to get feedback on whether you want this, if this is going in the right direction, if the public API is okay, etc. before I go too far.~~

Okay, well, I went on and finished up the Linux/X11, wasm, and Windows implementations anyway. There is a new `mouse_cursor` example you can try. Press `1234567890qw` to change the mouse cursor, `zx` to hide/show the mouse.

The interface looks like `ctx.set_mouse_cursor(CursorIcon::Pointer)`. The names in `CursorIcon` match the [CSS `cursor` property](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor) names.

Notes about the implementations

* Windows uses `LoadCursor`. This is the main limiting factor in what cursor icons are available (see [here](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-loadcursora) for a list); for example the "grabbing hand" requested in #171 is not on the list. AFAICT the way browsers get the "grabbing" hand is to embed their own .cur for it in the .exe.
* X11 uses `XCreateFontCursor`. The available cursors are listed [here](https://tronche.com/gui/x/xlib/appendix/b/). The grabbing hand isn't on this list. Neither are the nesw-resize/nwse-resize icons but I used  and  as approximations.

    An alternative is to use `XcursorLibraryLoadCursor` which would allow grabbing, nesw-resize, etc. However, it requires linking against `Xcursor` (or dlopening it or something). This works for me locally, but the CI machine doesn't have `Xcursor`, so I went with the simpler `XCreateFontCursor`.
* Wasm just sets the CSS `cursor` property on the canvas. This branch also makes `show_mouse` work for the first time on wasm.
* In X11 and wasm, where hiding the cursor is implemented by setting the cursor icon to an empty icon, miniquad needs to track the icon and whether the mouse is hidden so eg. (hide -> change icon -> show) will work correctly.
* I can't do darwin because I don't have a Mac.

Things to check

* Is the public API okay?
* The CursorIcon is passed to the backends by integer constants `SAPP_CURSOR_DEFAULT`, etc. repeated once in every backend. Is that the best way to do it?
* For X11/wasm, the variables to track the icon/visibility are globals. Should they be fields of sapp_state instead or something?

### DIFF
--- a/examples/mouse_cursor.rs
+++ b/examples/mouse_cursor.rs
@@ -1,0 +1,47 @@
+use miniquad::*;
+
+struct Stage {
+    inited: bool,
+}
+
+impl EventHandler for Stage {
+    fn update(&mut self, ctx: &mut Context) {
+        if !self.inited {
+            ctx.set_mouse_cursor(CursorIcon::Crosshair);
+            self.inited = true;
+        }
+    }
+
+    fn draw(&mut self, _ctx: &mut Context) {}
+
+    fn char_event(&mut self, ctx: &mut Context, character: char, _: KeyMods, _: bool) {
+        match character {
+            'z' => ctx.show_mouse(false),
+            'x' => ctx.show_mouse(true),
+            _ => (),
+        }
+
+        let icon = match character {
+            '1' => CursorIcon::Default,
+            '2' => CursorIcon::Help,
+            '3' => CursorIcon::Pointer,
+            '4' => CursorIcon::Wait,
+            '5' => CursorIcon::Crosshair,
+            '6' => CursorIcon::Text,
+            '7' => CursorIcon::Move,
+            '8' => CursorIcon::NotAllowed,
+            '9' => CursorIcon::EWResize,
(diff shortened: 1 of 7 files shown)


not-fl3:
Wow, thanks, great work done here! 

> The CursorIcon is passed to the backends by integer constants SAPP_CURSOR_DEFAULT, etc. repeated once in every backend. Is that the best way to do it?

I believe it is. You made a nice enum for the public API, so the user will not be affected by this and eventually we may came up with some better solution. But I personally think it is totally fine as it is :) 

> For X11/wasm, the variables to track the icon/visibility are globals. Should they be fields of sapp_state instead or something?

I also think this is fine, some other data started to leak out of sapp_state anyway. Maybe will refactor later.

PR #174, opened 2021-02-07
Update sokol_app.h for MacOS

Deseteral:
Updated `sokol_app.h` in MacOS backend to bring it in line with the most the recent version in sokol repo.

I've made some changes compared to the code found in sokol repo:
- added (yet) unused `SAPP_EVENTTYPE_RAW_DEVICE` that was necessary for the code to compile
- renamed `sapp_lock_mouse` function to `sapp_set_cursor_grab` (`sapp_set_cursor_grab` is what the rest of miniquad is using)

`SAPP_EVENTTYPE_RAW_DEVICE` is unused for now, but it will be useful for raw mouse input that I already have working on my fork. It needs this new version of sokol_app.h, so I will open PR with it after (and if) this one merges.

All of the miniquad and macroquad examples work correctly after this change. I'm also running it with my game, that I'm developing primarily on a Mac with no issues.

What this change most notably fixes is weird scaling issue on Macs with Retina displays. "Fixes" is a guess, because I don't really know how it should work. This is what I feel is correct behaviour, but please - correct me if I'm wrong.

To better understand what I'm talking about:

Before updating sokol_app.h, macroquad's "tree" example on my Retina display looked like this:
Screenshot 2021-02-07 at 15 42 32

This is the same example after I resize the window (even by 1px):
Screenshot 2021-02-07 at 15 42 44

The second screen also represents how it looks when I run it on my 2560x1440 monitor.
After updating sokol_app.h it looks like this no matter what display it's running on without any resizing tricks.
Please note that this is with `high_dpi` flag set to `false`.

Let me know if you want some changes to made - cleaning autogenerated sokol_app_darwin.rs for example :)


not-fl3:
> What this change most notably fixes is weird scaling issue on Macs with Retina displays. "Fixes" is a guess, because I don't really know how it should work. This is what I feel is correct behaviour, but please - correct me if I'm wrong.

Maybe it is possible to update only code responsible for this specific feature? At least we would know exactly what got changed.

I can't really do it myself without a mac/higdpi screen, so if nobody would volunteer on this - let's just merge it as is.

And lets keep it open for a little, maybe someone with mac will point to some issues

Deseteral:
> Maybe it is possible to update only code responsible for this specific feature? At least we would know exactly what got changed.

We could do that - the lack of meaningful diff is not useful. The problem is that I don't know what fixes this issue.

On the other hand this change also fixes some stuff that makes cursor lock working, so maybe other stuff that could be useful in the future is in here too. This update would also make future updates easier too (just dropping fresh sokol_app.h into the repo and running bindgen).

But if you insist on merging just the scaling/high DPI stuff just let me know and then I will try to work on that :)

not-fl3:
I will not do it myself - it is hard to do without a mac. 
So I'll keep this open for a little and if no one will volunteer on splitting this PR into small PRs that just fixes certain problems - will merge as is :) 

Mac is the last non-mobile platform that use sokol headers, and generally, I would like to eventually rewrite it in rust just as on windows/linux/web. Buuuut mac problem, again :D

Deseteral:
Hey! Just checked out master build of miniquad and it seems that this issue is now resolved. This PR can be closed :)

PR #175, opened 2021-02-08
try and create webgl2 context with webgl1 as fallback

heilmoli:
* I propose to take advantage of webgl2 where available (shadertoy does this too).  I think this makes sense as there are feature differences between platforms anyway. 
* Enabled OES_standard_derivatives by default (still needs to be enabled in the shader itself)
* Fixed a bunch of unexpected type coercion warnings in gl.js.


not-fl3:
Hi, thanks for PR, great work!
 
Just opened an issue to discuss how we can make it possible: https://github.com/not-fl3/miniquad/issues/176

PR #182, opened 2021-03-09
sapp-wasm: An attempt to fix high-dpi on web

not-fl3:


not-fl3:
Thanks for input @emilk @optozorax , merged, `v0.3.0-alpha.27` published

PR #183, opened 2021-03-10
graphics: add method to apply uniforms from u8 slice

optozorax:
This is needed in macroquad to use a dynamic-size array that stores uniforms.

### DIFF
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -930,18 +930,17 @@ impl Context {
     }
 
     pub fn apply_uniforms<U>(&mut self, uniforms: &U) {
-        self.apply_uniforms_buffer(unsafe {
-            std::slice::from_raw_parts(std::mem::transmute(uniforms), std::mem::size_of::<U>())
-        })
+        self.apply_uniforms_from_bytes(uniforms as *const _ as *const u8, std::mem::size_of::<U>())
     }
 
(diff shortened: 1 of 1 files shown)


src/graphics.rs
```diff
@@ -930,22 +930,30 @@ impl Context {
     }
 
     pub fn apply_uniforms<U>(&mut self, uniforms: &U) {
+        self.apply_uniforms_buffer(unsafe {
+            std::slice::from_raw_parts(std::mem::transmute(uniforms), std::mem::size_of::<U>())
```
optozorax, line 934: I am concerned about this line. Is everything works now like it works before? Is this transmute safe?

src/graphics.rs
```diff
@@ -930,22 +930,30 @@ impl Context {
...
+        })
+    }
+
+    pub fn apply_uniforms_buffer(&mut self, uniforms: &[u8]) {
```
optozorax, line 938: Is this name nice? Maybe `apply_uniforms_from_slice`?

optozorax: Renamed to `apply_uniforms_from_bytes`

src/graphics.rs
optozorax: Transmute removed.

PR #184, opened 2021-03-11
removes duplicate constant

nikita-skobov:
I compiled from source and noticed there was a duplicate GL_TIME_ELAPSED constant, so I removed it


nikita-skobov:
The one that is not removed is on line 922:

https://github.com/not-fl3/miniquad/blob/95770c7190ffe387d1fed1410fa374f6527f6d96/native/sapp-dummy/src/gl.rs#L922

not-fl3:
Thanks!

PR #188, opened 2021-03-21
android ndk glue code rewrite

nikita-skobov:
This is still a work in progress.

This is the first part of the proposed rewrite of the android glue code. The changes can be summarized by:

- adds dependency to the `sapp_android` package on `ndk`, `ndk-sys`, and `ndk-glue`.
- `sapp_android` provides an `init` function that can be specified by the `ndk_glue::main()` macro via: `#[cfg_attr(target_os = "android", ndk_glue::main(ndk_glue = "::miniquad::sapp_android"))]`
- our `init` function calls the sokol glue code that sets up android stuff for us.
- the sokol code implicitly calls a function called `sokol_main` so we also define one here which will call the users real `main` function.
- we provide an optional (but default included) macro to be used via: `#[miniquad::main]` which simply adds the `#[cfg_attr...]` for the ndk glue code mentioned above. basically this is a convenience wrapper that lets peoples projects be easily compiled for both android and non-android targets simply.
- adds example to Cargo.toml on how they could have multiple targets for android and non-android builds. ie: android builds would be required to be a `cdylib`, so theres a seperate example target that points to the same file. so if user wants to build for android they would do: `cargo apk build --example quad-android`, or otherwise to build for anything else: `cargo build --example quad`

**This PR is a work in progress and it doesnt actually include the rewrite of the sokol glue code like we talked about. Instead this is the kind of foundation/boilerplate of what I imagine that rewrite would look like.** So if we like what this looks like, then the next step would be to actually re-implement the code in `sapp-android/external/sokol_app.h` in rust, but I wanted to check in with the current progress before I start on that

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,8 @@ sapp-ios = { path = "./native/sapp-ios", version = "=0.1.2" }
 
 [target.'cfg(target_os = "android")'.dependencies]
 sapp-android = { path = "./native/sapp-android", version = "0.1.5" }
+ndk = { path = "../android-ndk-rs/ndk" }
+ndk-glue = { path = "../android-ndk-rs/ndk-glue" }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 sapp-wasm = { path ="./native/sapp-wasm", version = "=0.1.22" }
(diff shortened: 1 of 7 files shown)


nikita-skobov:
The reason the build fails for android is because the github actions file is using the old `notfl3/cargo-apk` docker image, which like I mentioned is a bit out of date. It should work with an up-to-date `cargo apk`. 

I am not sure why its failing for the macos builds, I dont have a mac so I cant test that unfortunately. I wonder if its something with the `cdylib` in the example, and maybe its trying to build that `quad-android` example even though that example is only meant to be built for android?

not-fl3:
fixed with https://github.com/not-fl3/miniquad/pull/207

PR #190, opened 2021-03-22
Cross-platform link opening api

optozorax:
Fixes https://github.com/not-fl3/egui-miniquad/blob/master/src/lib.rs#L166

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,9 @@ sapp-wasm = { path ="./native/sapp-wasm", version = "=0.1.22" }
 [target.'cfg(not(any(target_os="linux", target_os="macos", target_os="android", target_os="ios", target_arch="wasm32", windows)))'.dependencies]
 sapp-dummy = { path ="./native/sapp-dummy", version = "=0.1.5" }
 
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+webbrowser = "0.5.5"
+
 [dev-dependencies]
 glam = {version = "0.8", features = ["scalar-math"] }
 quad-rand = "0.1"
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -44,6 +44,9 @@ sapp-wasm = { path ="./native/sapp-wasm", version = "=0.1.22" }
 [target.'cfg(not(any(target_os="linux", target_os="macos", target_os="android", target_os="ios", target_arch="wasm32", windows)))'.dependencies]
 sapp-dummy = { path ="./native/sapp-dummy", version = "=0.1.5" }
 
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+webbrowser = "0.5.5"
```
optozorax, line 48: I am very concerned about this line. `miniquad` has no dependencies at all. Maybe this dependency should be under a feature? Or maybe remove this feature and make `link::open` working only on wasm? But, on other hand, this library has minimum dependencies too and works on Linux, Windows, MacOS, Android (!).

What do you think?

src/link.rs
```diff
@@ -0,0 +1,23 @@
+//! Functions to open links
+
+/// If not wasm, then open link in browser. If target is wasm, then link can be opened in the same tab, or in a new tab. But when link is opened in a new tab, browser may block it and ask user permission to do it.
+pub fn open(url: &str, new_tab: bool) {
```
optozorax, line 4: `new_tab` is needed because `egui` will have it in upcoming version: https://github.com/emilk/egui/issues/231.

not-fl3:
Looks like this functionality deserves a separate crate instead of being miniquad part.

optozorax:
Ok, I will move this out as a crate + plugin.

PR #191, opened 2021-03-23
docs: add page about plugins

optozorax:


ozkriff:
Maybe https://lib.rs/quad-rand is worth mentioning too?

not-fl3:
While we are here - maybe move stuff from wiki to /docs as well?

optozorax:
@ozkriff `quad-rand` is not a plugin, it has no `js`, therefore it's worth mentioning, but not in `List of plugins`.

@not-fl3 ok, I will add docs from the wiki as part of this PR.

PR #192, opened 2021-03-23
fix macos resize issue

extrawurst:
see https://github.com/not-fl3/macroquad/issues/177
and https://github.com/not-fl3/miniquad/pull/174


extrawurst:
this is a back port from upstream sokol header: https://github.com/floooh/sokol/blob/master/sokol_app.h#L2849

PR #199, opened 2021-04-21
Fix gl.js for iOS

python273:
`canvas.requestPointerLock` and `document.exitPointerLock` are not supported in Safari on iOS

Screenshot 2021-04-21 at 06 32 28


not-fl3:
Thanks for PR, good catch! 
I added a little comment to document why we use the empty function as a fallback

PR #202, opened 2021-04-29
graphics: fix zero-terminated string in the shader error

optozorax:
On Chrome, for unknown reason, string with one more zero-terminator is returned. This can lead problems when trying to panic with shader errors: panic can't be done because there will be redundant zero-terminators in the string, and CString can't be created. On Firefox there is no redundant zero-terminators.


src/graphics.rs
```diff
@@ -1189,10 +1189,15 @@ pub fn load_shader(shader_type: GLenum, source: &str) -> Result<GLuint, ShaderEr
             );
 
             assert!(max_length >= 1);
-            let error_message =
+            let mut error_message =
                 std::string::String::from_utf8_lossy(&error_message[0..max_length as usize - 1])
                     .to_string();
 
+            // On Wasm + Chrome, for unknown reason, string with zero-terminator is returned. On Firefox there is no zero-terminators in JavaScript string.
```
not-fl3, line 1196: :+1:

PR #203, opened 2021-05-03
add `#[derive(VertexLayout)]` that adds `VERTEX_ATTRIBUTES`

toyboot4e:
`#[derive(VertexLayout)]` generates `VERTEX_ATTRIBUTES` constant for limited types of fields.

## Background

I'm using Sokol and made such a macro. I thought it could be ported to `miniquad`.
Do you like it to be added?

## Example

```rust
use miniquad::VertexLayout;

#[test]
fn test() {
    // `#[derive(VertexLayout)]` panics at compile time if we miss `#[repr(C)]`
    #[repr(C)]
    #[derive(Debug, Clone, PartialEq, VertexLayout)]
    pub struct Vertex {
        pos: [f32; 2],
        color: [u8; 4],
    }

    assert_eq!(
        Vertex::VERTEX_ATTRIBUTES,
        &[
            miniquad::graphics::VertexAttribute::new(
                "pos",
                miniquad::graphics::VertexFormat::Float2,
            ),
            miniquad::graphics::VertexAttribute::new(
                "color",
                miniquad::graphics::VertexFormat::Byte4,
            ),
        ]
    );
}
````

### DIFF
--- a/miniquad-derive/Cargo.toml
+++ b/miniquad-derive/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "miniquad-derive"
+version = "0.1.0"
+edition = "2018"
+
+[lib]
+proc-macro = true
+
+[dependencies]
+proc-macro2 = "1.0.26"
+# need `extra-traits` feature for `Attribute` to implement `PartialEq`
+syn = { version = "1.0.70", features = ["derive", "parsing", "printing", "clone-impls", "proc-macro", "extra-traits"] }
+quote = "1.0.9"
+
+[dev-dependencies]
+miniquad = { path = "../" }
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
-/target
+**/target
 **/*.rs.bk
 Cargo.lock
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: miniquad-derive/Cargo.toml [hunk 1]
```diff
@@ -0,0 +1,16 @@
...
+version = "0.1.0"
+edition = "2018"
+
+[lib]
+proc-macro = true
+
+[dependencies]
+proc-macro2 = "1.0.26"
+# need `extra-traits` feature for `Attribute` to implement `PartialEq`
+syn = { version = "1.0.70", features = ["derive", "parsing", "printing", "clone-impls", "proc-macro", "extra-traits"] }
```
not-fl3, line 12: Using `syn` will triple miniquad's compilation time. I usually try to use proc-macro dependencies that do not use `syn`, like nanoserde. 

Maybe we can move this derive macro a third-party crate instead being a part of miniquad?

toyboot4e: > Using syn will triple miniquad's compilation time.
That's too bad!

Issue #193, opened 2021-04-02
add set_window_size for linux


joseluis:
I noticed the work was already done for windows and wasm in #189. So I'm opening the issue for tracking purposes bc I'm very interested.

Ideally it would be perfect if we could also: a) disable resizing, and b) stablish min and max limits for width and height.

rohan686:
I think this can be closed now

not-fl3:
It looks like `set_window_size` function is still unimplemented on linux and max

Bombfuse:
@eboatwright MacOS should work now btw, only linux remains unimplemented

eboatwright:
Thanks, but the only thing is now I have switched to Linux because my computer is from 2012 and was being discontinued XD

halfhorst:
This does _not_ work in MacOS now, as of Ventura 13.1

joseluis:
It looks like it's implemented now, closing.

PR #205, opened 2021-05-05
sapp-linux: implemented window_resizable flag #193

tharinduc:
This PR adds the `window_resizable` functionality for linux OS.

### DIFF
--- a/native/sapp-linux/src/lib.rs
+++ b/native/sapp-linux/src/lib.rs
@@ -249,6 +249,7 @@ pub struct sapp_desc {
         Option<unsafe extern "C" fn(_: *const libc::c_char, _: *mut libc::c_void) -> ()>,
     pub width: libc::c_int,
     pub height: libc::c_int,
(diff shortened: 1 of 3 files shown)


native/sapp-linux/src/lib.rs
```diff
@@ -473,8 +474,14 @@ pub unsafe extern "C" fn _sapp_x11_create_window(mut visual: *mut Visual, mut de
         1 as libc::c_int,
     );
     let mut hints = XAllocSizeHints();
-    (*hints).flags |= PWinGravity;
+    (*hints).flags |= PWinGravity | PMinSize | PMaxSize;
```
not-fl3, line 477: Should `PMinSize | PMaxSize` be used only when `window_resizable == false`?

tharinduc: Yes you are right. Let me update this

not-fl3:
Awesome, thanks!

PR #210, opened 2021-05-29
Fix windows clipboard reading too far and sometimes segfaulting

gamma-delta:


gamma-delta:
Sorry I'm lousy at Github, do I need to close the PR or do you pull it?

Issue #136, opened 2020-09-14
[Bug] `high_dpi` flag on Windows not working after window resizing


optozorax:
I have scale 1.5 in system, so `dpi_scale` should return `1.5`.

I compiled with `stable-x86_64-pc-windows-gnu`  toolchain

Before resizing:

![изображение](https://user-images.githubusercontent.com/18114381/93073252-cfd6f380-f6ac-11ea-9be4-810761922055.png)

After resizing: 

![изображение](https://user-images.githubusercontent.com/18114381/93073279-d7969800-f6ac-11ea-8500-a7ff0fea3b30.png)

optozorax:
Strange line of code https://github.com/not-fl3/miniquad/blob/master/native/sapp-windows/external/sokol/sokol_app.h#L7310

I tried scale 2.0, but nothing changes, so this line is not blame of this bug.

elsid:
Current implementation has following mistakes: uses wrong library for function to get dpi and asserts with a wrong condition for the result of that function. https://github.com/not-fl3/miniquad/pull/212 fixes the issue.

PR #212, opened 2021-06-11
Fix DPI on windows (issue #136)

elsid:
`GetDpiForMonitor` is a part of shcore not user32 according to
https://docs.microsoft.com/en-us/windows/win32/api/shellscalingapi/nf-shellscalingapi-getdpiformonitor.

And it returns `S_OK` on success that is an alias to 0 as defined here:
https://docs.microsoft.com/en-us/windows/win32/seccrypto/common-hresult-values

Should resolve https://github.com/not-fl3/miniquad/issues/136


not-fl3:
Oh, wow, good catch, thanks!

PR #214, opened 2021-06-12
graphics: add `user_dpi` variable

optozorax:
This can be used to scale user interface.


optozorax:
In WASM right now DPI can be changed in real-time by changing page zoom. I think this is convenient to have such variable that can change result of `dpi_scale` function on other platforms, like Android.

This variable should be inside of `miniquad`, and not on the user side, because there are some libraries that uses result of `dpi_scale()` function (like `egui`).

mkhan45:
Is there any way to change UI scale in egui_macroquad?

PR #224, opened 2021-07-31
sapp-wasm: fix clipboard miss chars

optozorax:
Previously byte length of pasted data was counted as count of char, which is wrong for UTF8. Now this is fixed by calculating exact count of needed bytes.


not-fl3:
thx!

PR #225, opened 2021-08-02
sapp-linux: read from clipboard into dynamic array

optozorax:
Previously linux clipboard could get only 32768 bytes, but now it can get any number of bytes using dynamic array.


not-fl3:
:+1:

PR #226, opened 2021-08-11
implementing set_fullscreen for linux

ricardoalcantara:
I think this might be useful, it's based on sokol version the linux set_fullscreen.

I could also upload the example file.


Bombfuse:
Would we be able to merge this soon? set_fullscreen on linux would be huge gains

not-fl3:
merged :) https://github.com/not-fl3/miniquad/commit/041b6fd01374488cc47b9da45580e8d94a334344

PR #229, opened 2021-08-28
implemented sapp_dpi_scale

avenger0815:
implemented fn sapp_dpi_scale to avoid the panic when running into the function (happens to me when using macroquad/egui).


not-fl3:
Thanks!

PR #245, opened 2021-11-01
native/sapp-wasm: key_up event includes modifiers

urholaukkarinen:
key_up event in wasm implementation did not include modifiers. This caused the application to lose modifier information when any key was released.

How to reproduce:
In a macroquad wasm application, while holding ctrl key, press and release alt key.
After releasing the alt key, macroquad thinks ctrl is also no longer pressed.


not-fl3:
Nice, thanks for PR!

PR #246, opened 2021-11-06
Support BYTE, SHORT, INT type of index buffer

smokku:
This change adds configurable type for index buffer index type.

Supports all types supported by OpenGL: `GL_UNSIGNED_BYTE`, `GL_UNSIGNED_SHORT`, `GL_UNSIGNED_INT`.

https://khronos.org/registry/OpenGL-Refpages/gl4/html/glDrawElementsInstanced.xhtml

### DIFF
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -432,9 +432,11 @@ struct CachedAttribute {
 
 struct GlCache {
     stored_index_buffer: GLuint,
+    stored_index_type: Option<IndexType>,
     stored_vertex_buffer: GLuint,
     stored_texture: GLuint,
     index_buffer: GLuint,
+    index_type: Option<IndexType>,
     vertex_buffer: GLuint,
     textures: [GLuint; MAX_SHADERSTAGE_IMAGES],
     cur_pipeline: Option<Pipeline>,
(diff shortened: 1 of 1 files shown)


src/graphics.rs
```diff
@@ -470,18 +473,19 @@ impl GlCache {
...
             self.stored_index_buffer = self.index_buffer;
+            self.stored_index_type = self.index_type;
         }
     }
 
     fn restore_buffer_binding(&mut self, target: GLenum) {
         if target == GL_ARRAY_BUFFER {
-            if self.stored_vertex_buffer != 0 {
-                self.bind_buffer(target, self.stored_vertex_buffer);
+            if self.stored_vertex_buffer != 0 && self.stored_vertex_buffer != target {
```
not-fl3, line 482: Hmm, I did not quite get that new check. What is its purpose?

src/graphics.rs
```diff
@@ -1704,19 +1752,24 @@ impl Buffer {
             gl_buf,
             buffer_type,
             size,
+            index_type: None,
         }
     }
 
-    pub fn update<T>(&self, ctx: &mut Context, data: &[T]) {
+    pub fn update<T>(&mut self, ctx: &mut Context, data: &[T]) {
```
not-fl3, line 1759: Unfortunately `update` can't have `&mut self`. 

Buffer is a copy-able handle for OpenGl buffer. Modifying one of the copies will not necessary modify the one that will be lately used for drawing :( 

The idea here - we can set up some fields in a Buffer struct during the initialization, but later on, the Buffer struct is just an ID

src/graphics.rs
smokku: When you update the buffer with new IndexType it will store current index_type `None`, update it and then restore with stored `None`.

PR #248, opened 2021-11-27
Fix windows set_window_size

Bombfuse:
Discovered the windows resizing was broken, turns out setting window size includes the header so we need to recalculate the desired windows size from the client rect :/


not-fl3:
Thanks!

PR #259, opened 2022-01-16
Make use of high_dpi flag

samoylovfp:


jackfranklin:
I have confirmed for me on a PopOS device on a 4k screen that this PR fixes the problem; with this change I can set high_dpi to true/false and have the rendering change as I would expect it to.

PR #272, opened 2022-03-10
native/sapp-wasm: Use relative position for touch events

dragly:
Previously, touch events were not registered relative to their canvas,
but to the document. This means that touch events would only be
positioned correctly if the canvas was placed in the upper left
corner of the document.

This change applies the same logic to touch events as is currently
applied to mouse events and thereby positions touch events correctly.

### DIFF
--- a/native/sapp-wasm/js/gl.js
+++ b/native/sapp-wasm/js/gl.js
@@ -1166,28 +1166,32 @@ var importObject = {
                 event.preventDefault();
 
                 for (const touch of event.changedTouches) {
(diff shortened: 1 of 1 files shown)


dragly:
Here is an example that uses this code (in place of mq_js_bundle.js). I discovered the issue when testing the blog post on a mobile device: https://dragly.org/2022/02/18/neuron/

not-fl3:
Thanks! :+1:

dragly:
No problem :) The file in macroquad called `mq_js_bundle.js` should probably also be updated, but I could not figure out exactly which files needed to be concatenated to do so correctly myself. I ended up getting errors about assigning to a non-existent variable called `js_objects`, for instance.

PR #275, opened 2022-04-17
File drag & drop support for Windows, Linux and WASM

urholaukkarinen:
Miniquad can now handle dropped files on Windows, Linux and WASM.

A file path on Windows may be at most 260 characters long (16 bits each). On Mac it is 1024 bytes and on Linux it is 4096 bytes. Is it okay to have a 4096 bytes long array in sapp_event? If PathBuf is used, the event can no longer be `Copy`.

In browsers you cannot read a file from the filesystem with a path, so the file bytes are read immediately in the file drop event. In native applications the file reading can be left to the client.

### DIFF
--- a/native/sapp-linux/src/clipboard.rs
+++ b/native/sapp-linux/src/clipboard.rs
@@ -13,19 +13,6 @@ const AnyPropertyType: libc::c_long = 0 as libc::c_long;
 
 type Time = libc::c_ulong;
 
-extern "C" {
-    pub fn XConvertSelection(
-        _: *mut Display,
-        _: Atom,
-        _: Atom,
-        _: Atom,
-        _: Window,
-        _: Time,
-    ) -> libc::c_int;
-
-    pub fn XSetSelectionOwner(_: *mut Display, _: Atom, _: Window, _: Time) -> libc::c_int;
-}
-
 pub unsafe fn get_clipboard(
     mut bufname: *const libc::c_char,
     mut fmtname: *const libc::c_char,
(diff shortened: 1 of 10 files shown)


urholaukkarinen:
looks like sokol has functions for getting dropped file paths and data, and the event itself does not have anything except the event type. That's probably better than having file path and data in the event.

urholaukkarinen:
I changed the implementation so that sapp_event no longer carries the file path or bytes. Currently the drag & drop feature is always enabled. I've tested this with modified egui-miniquad demo on Windows, Linux and web.

not-fl3:
Wow, impressive work, thank you!

I was thinking how to merge it for a little bit, sorry for delay :) 

I have some opinions on the API here, but, taking into account that all sapp-related stuff is being heavily refactored right now (yes, I still did not opened this PR...) - I would rather merge this as is and later on will change a few things. Hope this sounds good to you! 

Thanks again!

urholaukkarinen:
Sounds good to me! I tried to mimic the existing api a bit when implementing this, but I did also expect I would need to make some changes after reviews.

nokola:
This is awesome! Adding some info: on Windows you can have > 260 character paths:
1. Press the Windows key, type gpedit.msc and press Enter.
2. Navigate to Local Computer Policy > Computer Configuration > Administrative Templates > System > Filesystem > NTFS.
3. Double click the Enable NTFS long paths option and enable it.

urholaukkarinen:
@nokola very good point. Currently this implementation assumes the path is at most 260 characters, but it would be easy to fix it.

PR #276, opened 2022-05-03
README.md: update system dependencies for ubuntu

Yeicor:


not-fl3:
Hmm, does miniquad depend on libasound2 somehow?

Yeicor:
My bad, sorry.

I copied the dependencies command from miniquad instead of macroquad and it gave me the missing dependency error so I thought I would update it, but I was wrong.

PR #278, opened 2022-05-08
0.3 Stabilization

not-fl3:
# 0.3 Stabilization PR

Miniquad have been in alpha for way too long. This PR finish this. 

# API change:

No more EventHandler/EventHandlerFree traits.
Now just EventHandler.

```diff
     miniquad::start(conf::Conf::default(), |mut ctx| {
-        UserData::owning(Stage::new(&mut ctx), ctx)
+        Box::new(Stage::new(&mut ctx))
     });
```

`clipboard::set(ctx: &mut Context)` became an associated function and `clipboard` module got removed:

```diff
-    clipboard::set(ctx, data);
+    ctx.clipboard_set(data);
```

# sapp-* crates

Miniquad used to depend on sokol_app.h's C and ObjectiveC. To build sokol_app.h on various platforms different [build-dependencies] were required.
Back then(and probably still) cargo did not had per-platform [build-dependencies] section. So it was a natural solution to put a native code into a subcrates. Each could
have its own build-dependencies and do not interference with other platforms.

But lately, one by one, sapp-* got rewrited in rust.

And having them as separate crates became a huge pain - its 8 crates, each
with its own versions. A little bug fix required a version bump, etc, etc.

We do not have any [build-dependencies] anymore at all. And no C/ObjC code as
well.

# Linux

Linux used to be covered with sapp-linux, sapp-kms, sapp-linux from a wayland branch and sapp-linux from a egl branch.

For sapp-linux, by default, miniquad used to statically link (-lX11) with libX11, libGLX and libGL. But on some platforms there is X11, but not libGLX (some rapsberry pi distros). For this case we had a branch that statically linked with libEGL instead of libGLX.
Same for wayland - there was a branch with X11 replaced with wayland and lots of static linking.

Now all this work got together and may be used from a single binary! Miniquad can check if libX11 or wayland-client exist, dynamically load them and create the context. The only statically linked dependency is libc. In other words, the same binary can work in either pure X11 or pure Wayland environment.

```
#[derive(Debug)]
pub enum LinuxX11Gl {
    /// Use libGLX.so/libGLX.so.0 and its funciton for creating OpenGL context
    /// If there is no libGLX - just panic right away
    GLXOnly,
    /// Use libEGL.so/libEGL.so.0 and its funciton for creating OpenGL context
    /// If there is no libEGL - just panic right away
    EGLOnly,
    /// Use libGLX and if there is not libGLX - try libEGL.
    /// The default option.
    GLXWithEGLFallback,
    /// Use libEGL and if there is not libEGL - try libGLX.
    EGLWithGLXFallback,
}

#[derive(Debug)]
pub enum LinuxBackend {
    X11Only,
    WaylandOnly,
    X11WithWaylandFallback,
    WaylandWithX11Fallback,
}
```

# Linux, dynamic linking

Just to reiterate on this - miniquad do not depend on any *-dev packages. Miniquad dlopen .so files without static linking. It makes miniquad binaries very portable - if you have libc and at least some graphics environment - x11, wayland, kms, whatever - miniquad will find a way to create a context!

# MacOs

`sapp-darwin` was the last resort sokol-app.h. Miniquad used to call `cc-rs` to compile sokol-app's Objective C.

Not anymore. MacOs backend got reimplemented in Rust with the help of [objc  crate](https://crates.io/crates/objc). Miniquad's implementation is a blend between ideas and design from sokol-app and Makepad's macos implementation](https://github.com/makepad/makepad/blob/live/platform/src/platform/apple)

# GL loader

Linux, windows and android now share the same OpenGL loader, src/native/gl.rs


Cargo.toml
```diff
@@ -14,36 +14,25 @@ keywords = ["graphics", "3D", "opengl", "gamedev", "windowing"]
...
-sapp-android = { path = "./native/sapp-android", version = "=0.1.10" }
-
-[target.'cfg(target_arch = "wasm32")'.dependencies]
-sapp-wasm = { path ="./native/sapp-wasm", version = "=0.1.26" }
+libc = "0.2"
+ndk-sys = "0.2"
 
-[target.'cfg(not(any(target_os="linux", target_os="dragonfly", target_os="freebsd", target_os="netbsd", target_os="openbsd", target_os="macos", target_os="android", target_os="ios", target_arch="wasm32", windows)))'.dependencies]
-sapp-dummy = { path ="./native/sapp-dummy", version = "=0.1.5" }
+[target.'cfg(target_os = "macos")'.dependencies]
```
simlay, line 32: Unclear if you meant to remove iOS but you may have meant something like ` [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]`.

not-fl3: Thanks! 

Honestly I just never tested it on iOS, I am pretty sure there will be more issues there :(

Issue #280, opened 2022-05-11
lib.rs copy-pasted code issue?


dgollub:
Hi,

I see this in lib.rs

```
        match conf.platform.linux_backend {
            conf::LinuxBackend::X11Only => {
                native::linux_x11::run(&conf, f).expect("X11 backend failed")
            }
            conf::LinuxBackend::WaylandOnly => {
                native::linux_wayland::run(&conf, f).expect("Wayland backend failed")
            }
            conf::LinuxBackend::X11WithWaylandFallback => {
                if native::linux_x11::run(&conf, f).is_none() {
                    println!("Failed to initialize through X11! Trying wayland instead");
                    native::linux_wayland::run(&conf, f);
                }
            }
            conf::LinuxBackend::WaylandWithX11Fallback => {
                if native::linux_x11::run(&conf, f).is_none() {
                    println!("Failed to initialize through wayland! Trying X11 instead");
                    native::linux_wayland::run(&conf, f);
                }
            }
        }
```
Shouldn't the last match arm rather be this

```
            conf::LinuxBackend::WaylandWithX11Fallback => {
                if native::linux_wayland::run(&conf, f).is_none() {
                    println!("Failed to initialize through wayland! Trying X11 instead");
                    native::linux_x11::run(&conf, f);
                }
            }
```
?

-- Daniel

not-fl3:
Oh, sure, feel free to open a PR!

dgollub:
As requested I created a PR. Please let me know if anything is missing or needs changing. Thanks.

PR #281, opened 2022-05-11
Fix for Issue 280: lib.rs copy-pasted code issue

dgollub:
This fixes the code in lib.rs as described in issue #280


dgollub:
Hmmm, this PR doesn't look correct. Not sure why.

PR #284, opened 2022-05-15
Fix wasm file loading status handling

lunabunn:
Previously, when the `XHRHttpRequest` was successfully completed with a non-200 status code, `file_loaded` was never fired, causing the game to hang. This just calls `file_loaded` in such cases with a `null` buffer so that an `Err` can be returned correctly.

@nobbele reports that even with this fix, there is a screen flicker issue when the the file loading fails with a 404. Noting here because I have no idea what could be causing that or how to fix it and have neither the time nor the expertise to investigate further :P

### DIFF
--- a/js/gl.js
+++ b/js/gl.js
@@ -1269,7 +1269,10 @@ var importObject = {
 
                     FS.loaded_files[file_id] = uInt8Array;
                     wasm_exports.file_loaded(file_id);
-                }
+                } else {
+		    FS.loaded_files[file_id] = null;
+		    wasm_exports.file_loaded(file_id);
+		}
             }
             xhr.onerror = function (e) {
                 FS.loaded_files[file_id] = null;


js/gl.js
```diff
@@ -1269,7 +1269,10 @@ var importObject = {
 
                     FS.loaded_files[file_id] = uInt8Array;
                     wasm_exports.file_loaded(file_id);
-                }
+                } else {
```
not-fl3, line 1272: I am not exactly sure that this will work? Not really sure what to do with this PR, will convert to draft and leave in case someone might find it useful

not-fl3: Oh, apparently github do not allow converting to drafts.. than just closing for now

lunabunn: Why wouldn't it work? The current behavior is incorrect, do you have another idea for a fix?

not-fl3: Does it work? I did not really tried this, if it works fine - sure, lets get it merged

PR #286, opened 2022-05-15
Handle and return non-success status codes when loading files on web

lunabunn:
Successor of #284 

- Add `WebAssetLoadingError(u16)` to `fs::Error` a la `AndroidAssetLoadingError`
- Call file loaded callback even on failure (fixes infinite hang when file fails to load)
- Throw `fs::Error::DownloadFailed` on 404 (file not found), otherwise throw `fs::Error::WebAssetLoadingError` with the status code
- Treat errors as a 404 (file not found) and also throw `fs::Error::DownloadFailed`

Everything is tested, but I'm leaving this as a draft because:
- This is a breaking change; should the gl.js version be bumped to `0.1.27`?
- This may return garbage if the server responds with an invalid status code (such as negative status codes or status codes over `i16::MAX`); I think this is fine, but do you want me to special-case this and throw `DownloadFailed`?

### DIFF
--- a/js/gl.js
+++ b/js/gl.js
@@ -1269,7 +1269,10 @@ var importObject = {
 
                     FS.loaded_files[file_id] = uInt8Array;
                     wasm_exports.file_loaded(file_id);
-                }
+                } else {
+		    FS.loaded_files[file_id] = null;
+		    wasm_exports.file_loaded(file_id);
+		}
             }
             xhr.onerror = function (e) {
                 FS.loaded_files[file_id] = null;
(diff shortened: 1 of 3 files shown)


lunabunn:
@not-fl3 I already asked on Discord a few months ago, but can I please get your input on this? I have un-drafted it in case that's what was stopping you.

I see that on master you have a different solution for this, but the JS code there does not gracefully handle non-success error codes and is not very idiomatic overall. I think this is a nicer approach.

As stated in the PR itself, this is mostly ready to merge. Just waiting for review.

not-fl3:
Thanks for pr! Will look closely into this after 0.4 merge!

PR #289, opened 2022-05-29
Fix mouse buttons in Wasm

maciejhirsz:
JS still thinks we are in sapp land:

```rust
pub const sapp_mousebutton_SAPP_MOUSEBUTTON_MIDDLE: sapp_mousebutton = 2;
pub const sapp_mousebutton_SAPP_MOUSEBUTTON_RIGHT: sapp_mousebutton = 1;
pub const sapp_mousebutton_SAPP_MOUSEBUTTON_LEFT: sapp_mousebutton = 0;
pub const sapp_mousebutton_SAPP_MOUSEBUTTON_INVALID: sapp_mousebutton = -1;
```

Current code in Rust is:
https://github.com/not-fl3/miniquad/blob/5fe44e7146b88c34c0ae04cae62f003d8992e247/src/native/wasm/keycodes.rs#L7-L14

Bonus: minor thing I just spotted at random (`into_owned` avoids allocating if `Cow` was already owned).

### DIFF
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -1231,10 +1231,10 @@ pub fn load_shader(shader_type: GLenum, source: &str) -> Result<GLuint, ShaderEr
             assert!(max_length >= 1);
             let mut error_message =
                 std::string::String::from_utf8_lossy(&error_message[0..max_length as usize - 1])
-                    .to_string();
+                    .into_owned();
 
             // On Wasm + Chrome, for unknown reason, string with zero-terminator is returned. On Firefox there is no zero-terminators in JavaScript string.
-            if error_message.chars().last() == Some('\0') {
+            if error_message.ends_with('\0') {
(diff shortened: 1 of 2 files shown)


js/gl.js
```diff
@@ -424,10 +424,10 @@ const SAPP_MODIFIER_SUPER = 8;
 
 function into_sapp_mousebutton(btn) {
     switch (btn) {
-        case 0: return 0;
```
not-fl3, line 427: Hmm, maybe its easier to fix this on the rust side? While technically all gl.js's should be versioned and updated, I am pretty sure lots of people just use the last one and this will break existing code quite a lot :(

maciejhirsz: Not sure about easier since that will touch all other platforms (I think?), but I think you're right in that rust updates don't require also getting a new .js file every time, will get on it!

PR #294, opened 2022-06-10
MacOS: Activate app on startup

maciejhirsz:
Forces the app to the foreground so the window shows up after `cargo run` without having to click it in the task bar.

This should be the last one of these micro PRs for MacOS for a while I hope.


not-fl3:
> This should be the last one of these micro PRs for MacOS for a while I hope.

All those little things make a big difference, thank you very much for your work!

PR #296, opened 2022-06-17
Remove <'a, 'b> from Context definition

not-fl3:
Pre `0.3` `EventHandler` looked like this:
```rust
pub trait EventHandler {
    fn update(&mut self, _ctx: &mut Context);
    ..
}
```

Post `0.3` looks exactly the same: 
```rust
pub trait EventHandler {
    fn update(&mut self, _ctx: &mut Context);
    ..
}
```

So I thought this is going work and no breaking change happened. What a mistake! 
I the new miniquad Context got a lifetime, so it was actually this:
`fn update<'a>(&'a mut self, _ctx: &'a Context<'a, 'a>)`

The new context was basically a tuple struct with two references. This allowed some nice type-safe things to happen.

The problem with lifetimes in the context - it is really painful to keep the reference to the context inside user code. Especially in the case of `&mut Context` -> `&mut dyn Any`. 

This PR removes the lifetimes. 

How it works:

Most backends have something like:
```rust
struct WindowPayload {
  context: GraphicsContext,
  display: NativeDisplayImplementation
}
```

When its time to call some EventHandler function, miniquad used to do basically this:
```rust
struct Context(&mut GraphicsContext, &mut dyn NativeDisplay);
```

```rust
event_handler.update(&mut Context::new(&mut payload.context, &mut payload.display));
```

display became a trait-object, memory/aliasing safety is doing well, a function called. 

After this PR miniquad is doing this:

```rust
event_handler.update(payload.context.as_mut(&mut payload.display));
```

I believe this is UB-free and safe enough until event-handler's calls are using `context.as_mut(&mut display)` and not `&mut context`. I would love to enforce `as_mut` call on compile time, but, unfortunately, it doesn't seem possible without either introducing some sort of a wrapper type around Context with a lifetime or dealing with raw pointers everywhere :(

### DIFF
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -302,8 +302,8 @@ impl Shader {
         meta: ShaderMeta,
     ) -> Result<Shader, ShaderError> {
         let shader = load_shader_internal(vertex_shader, fragment_shader, meta)?;
-        ctx.a.shaders.push(shader);
-        Ok(Shader(ctx.a.shaders.len() - 1))
+        ctx.shaders.push(shader);
+        Ok(Shader(ctx.shaders.len() - 1))
     }
 }
 
@@ -595,7 +595,7 @@ impl RenderPass {
                     0,
                 );
             }
-            glBindFramebuffer(GL_FRAMEBUFFER, ctx.a.default_framebuffer);
+            glBindFramebuffer(GL_FRAMEBUFFER, ctx.default_framebuffer);
         }
         let pass = RenderPassInternal {
             gl_fb,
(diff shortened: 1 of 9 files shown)


not-fl3:
Was thinking about this a lot

I would prefer to keep GraphicsContext and dyn NativeDisplay completely separated, maybe like this:
```
fn update(&mut self, to_draw_things: &mut GraphicsContext, to_manipulate_the_window: &mut dyn NativeDisplay)
```

But at the same time I believe that one single context is sort of hard requirement for miniquad 0.3

It should be possible to get `&mut Context` without any lifetimes in the Context, and this `&mut Context` should be enough to manipulate the window and do rendering. Lots of libraries depend on this!

So while I do not particularly like this PR, I am going to merge this. If something will be broken - it's a bug, but the design should keep the same from now on.

PR #297, opened 2022-06-20
Fix handling of repeated keycodes in linux_x11.

wohe:
A typo prevented keys to be properly considered released for the purpose of the repeat flag.
This caused only the very first key press of a key to be reported as non-repeated, e.g.
breaking macroquad's is_key_pressed on Linux.

### DIFF
--- a/src/native/linux_x11.rs
+++ b/src/native/linux_x11.rs
@@ -486,7 +486,7 @@ impl X11Display {
             3 => {
                 let keycode = (*event).xkey.keycode;
                 let key = self.translate_key(keycode as _);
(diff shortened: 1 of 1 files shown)


src/native/linux_x11.rs
```diff
@@ -486,7 +486,7 @@ impl X11Display {
             3 => {
                 let keycode = (*event).xkey.keycode;
                 let key = self.translate_key(keycode as _);
-                self.repeated_keycodes[(keycode & 0xffe) as usize] = false;
```
not-fl3, line 489: ohhh

good catch, thank you!

PR #298, opened 2022-06-20
RPi Compilation Fixed

iamwacko:
These changes let miniquad compile on  Raspberry Pi's.

### DIFF
--- a/src/native/linux_wayland.rs
+++ b/src/native/linux_wayland.rs
@@ -140,18 +140,108 @@ unsafe extern "C" fn seat_handle_capabilities(
 ) {
 }
 
+#[cfg(not(all(any(target_arch = "x86", target_arch = "x86_64"), target_os = "linux")))]
 extern "C" fn seat_handle_name(
     _data: *mut std::ffi::c_void,
     _seat: *mut wl_seat,
-    _name: *const _,
+    _name: *const u8,
 ) {
 }
 
+#[cfg(all(any(target_arch = "x86", target_arch = "x86_64"), target_os = "linux"))]
+extern "C" fn seat_handle_name(
+    _data: *mut std::ffi::c_void,
+    _seat: *mut wl_seat,
+    _name: *const u8,
+) {
+}
+
+#[cfg(not(all(any(target_arch = "x86", target_arch = "x86_64"), target_os = "linux")))]
 unsafe extern "C" fn registry_add_object(
     data: *mut std::ffi::c_void,
     registry: *mut wl_registry,
     name: u32,
-    interface: *const _,
+    interface: *const u8,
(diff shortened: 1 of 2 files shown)


src/native/linux_wayland/extensions.rs
```diff
@@ -58,7 +58,7 @@ macro_rules! wayland_interface {
             }),*];
 
             pub static mut $name: wl_interface = wl_interface {
-                name: concat!(stringify!($struct_name), '\0').as_ptr() as *const i8,
+                name: concat!(stringify!($struct_name), '\0').as_ptr() as *const u8,
```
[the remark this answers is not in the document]
iamwacko: Thanks

iamwacko:
It's a bit ugly, but the cfg macro should solve the problem.

src/native/linux_wayland.rs
```diff
@@ -140,18 +140,108 @@ unsafe extern "C" fn seat_handle_capabilities(
 ) {
 }
 
+#[cfg(not(all(any(target_arch = "x86", target_arch = "x86_64"), target_os = "linux")))]
```
not-fl3, line 143: is it really a necessity? Would be nice to keep wayland support on non-x86 targets

iamwacko: Support should still be there. cfg allows you to define two functions in the code but only compile one, depending on the platform.. That is what I did here, for both of the problematic functions.

iamwacko: @not-fl3 You should be able to use u8 on all platforms, that is what the [docs](https://doc.rust-lang.org/std/ffi/struct.CStr.html) say, but the error (shown below) says otherwise. Perhaps this is a problem with the Rust std library?

image

not-fl3: yeah, I do not really understand why its i8/u8

but it kinda makes sense: 
```
It can be constructed safely from a &[[u8](https://doc.rust-lang.org/std/primitive.u8.html)] slice, or unsafely from a raw *const c_char
```

we are using unsafe `from_ptr` that accepts `c_char`

not-fl3: and from the practical point of view the only important part of the doc is `from_ptr` signature: 

```rust
pub unsafe fn from_ptr<'a>(ptr: *const c_char) -> &'a CStr
```

its a `c_char`. `c_char` is either i8 or u8: https://doc.rust-lang.org/std/os/raw/type.c_char.html

so we should just ensure that its `c_char` in our export "C" functions

not-fl3: Could you update the PR with the `c_char` (basically with the diff a few comments above)? I'll merge it right away :)

PR #303, opened 2022-07-10
Derive Clone on ShaderMeta and a few other types

Jaso55555:
All the examples seem to run fine with this addition, and it helps when two shaders need the same ShaderMeta.


not-fl3:
Thanks for PR!

PR #311, opened 2022-09-01
Put example shaders in separate files

Speykious:
This PR just takes the raw shader strings of the examples and puts them in separate files.
The `include_str!` macro makes sure that the file exists at compile time.

Some shaders were exactly the same so I tried to give them a sensible name so that they get reused.
Let me know what you think!


Speykious:
¯\_(ツ)_/¯

PR #312, opened 2022-09-04
Full fullscreen on Android

Speykious:
This PR makes the app truly fullscreen on Android. It covers the entire area of the phone including the upper band that could be partially hidden on phones with wider screens.

One thing I'm not sure is if this should be toggleable separately from the `fullscreen` setting in Miniquad. Would some people want their fullscreen to not cover the partially hidden part of the screen on wider phones? I don't know.

## Before
(notice the black band on the left, where stuff like the front camera could be.)
image

## After
image

### DIFF
--- a/java/MainActivity.java
+++ b/java/MainActivity.java
@@ -10,6 +10,7 @@ import android.util.Log;
 import android.view.View;
 import android.view.Surface;
 import android.view.Window;
+import android.view.WindowManager.LayoutParams;
 import android.view.SurfaceView;
 import android.view.SurfaceHolder;
 import android.view.MotionEvent;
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks for PR!

andreymal:
I guess this change broke Android < 9 compatibility

https://github.com/not-fl3/miniquad/issues/328#issuecomment-1666944825

not-fl3:
> I guess this change broke Android < 9 compatibility
> 
> [#328 (comment)](https://github.com/not-fl3/miniquad/issues/328#issuecomment-1666944825)

Please, use new issues with a link to old PRs instead of bumping old PRs themselves! Its really hard to follow the whole conversation this way.

andreymal:
@not-fl3 bumping old PRs is a good thing because authors of there PRs may receive notifications

PR #313, opened 2022-09-05
map special key to hid code for linux x11 in order to get special key event

skydig:
to get special key event, I map special key to hid code according to file keylayouts.h as  attached
[keylayouts.zip](https://github.com/not-fl3/miniquad/files/9487905/keylayouts.zip)
code value is in unicode private area(0xE000-0xF8FF)


not-fl3:
oh, sorry, missed out on this

Thanks for PR!

PR #319, opened 2022-10-31
Implement clipboard for MacOS

### LINKED ISSUE #172: native/sapp-darwin: Implement mac clipboard
Entrypoint for this issue: https://github.com/not-fl3/miniquad/blob/master/src/clipboard.rs#L60

On MacOs clipboard is not implemented and is falling back to a `dummy` backend. 

Some tips on implementation: as far as I know upstream sokol_app have a clipboard implementation, so the plan 
- take a look on the miniquad's clipboard api and linux/windows impls as a reference
- do a lot of copy-pasting from upstream sokol_app to native/sapp-darwin's sokol_app
- write some rust in native/sapp-darwin to match the miniquad clipboard api

ollej:
A naive implementation of clipboard_set and clipboard_get for MacOS. Tested on MacOS 12.6.

Fixes #172

### DIFF
--- a/src/native/macos.rs
+++ b/src/native/macos.rs
@@ -89,7 +89,15 @@ impl crate::native::NativeDisplay for MacosDisplay {
     fn clipboard_get(&mut self) -> Option<String> {
         None
     }
-    fn clipboard_set(&mut self, _data: &str) {}
+    fn clipboard_set(&mut self, data: &str) {
+        let str: ObjcId = str_to_nsstring(data);
+        unsafe {
+            let pasteboard: ObjcId = msg_send![class!(NSPasteboard), generalPasteboard];
+            let () = msg_send![pasteboard, clearContents];
+            let arr: ObjcId = msg_send![class!(NSArray), arrayWithObject: str];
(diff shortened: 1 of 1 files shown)


src/native/macos.rs
```diff
@@ -87,9 +87,22 @@ impl crate::native::NativeDisplay for MacosDisplay {
         }
     }
     fn clipboard_get(&mut self) -> Option<String> {
-        None
+        unsafe {
+            let pasteboard: ObjcId = msg_send![class!(NSPasteboard), generalPasteboard];
+            let content: ObjcId = msg_send![pasteboard, stringForType: NSStringPboardType];
+            let string = nsstring_to_string(content);
+            (!string.is_empty()).then(|| string)
```
not-fl3, line 94: to be honest I would prefer something like
```
if string.is_empty() {
  return None;
}
string
```

ollej: I totally agree, I've pushed a fix for this.

not-fl3:
Thanks for PR!

PR #323, opened 2022-11-12
wasm: Fix crash when calling load_file with non-existant file

FaunKr:
This fixes: https://github.com/not-fl3/macroquad/issues/486


not-fl3:
Thanks for PR! I added a little note on why === 4 :)

PR #324, opened 2022-11-15
Windows raw mouse motion events and cursor grabbing

### LINKED ISSUE #139: raw_mouse_motion event does not work on Windows
As well as sapp_set_cursor_grab

plish-plash:
Implements raw mouse motion events for Windows. Also implements the set_cursor_grab and show_mouse functions.

- There was already a WM_EVENT handler for raw mouse motion events, however, it's necessary to call RegisterRawInputDevices to enable this message. I added this call right after window creation.
- Similarly, there was already a update_clip_rect function to confine the cursor to the window, but cursor_grabbed was always false, so this function was never called.

Tested on Windows 10.

Fixes #139

### DIFF
--- a/src/native/windows.rs
+++ b/src/native/windows.rs
@@ -7,6 +7,7 @@ use crate::{
 
 use winapi::{
     shared::{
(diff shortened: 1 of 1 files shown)


not-fl3:
Awesome, thanks for PR!

PR #325, opened 2022-11-17
Wasm  fix german "-" keycode

FaunKr:
It similiar problem as https://github.com/not-fl3/macroquad/issues/512. described in https://github.com/not-fl3/macroquad/issues/512#issuecomment-1318915843

Tested on German and US Layout


not-fl3:
thanks!

Issue #326, opened 2022-11-17
Windows high dpi borderless fullscreen


bananaturtlesandwich:
on windows when the config is set to have both fullscreen and high_dpi as true
```rust
miniquad::conf::Conf {
            window_title: ":)".to_string(),
            sample_count: 8,
            fullscreen: true,
            high_dpi: true,
            ..Default::default()
        }
```
the window will start up like this
<img width="960" alt="image" src="https://user-images.githubusercontent.com/71292624/202563081-5833275e-8e1e-4a82-a066-eb73bf6187f6.png">

bananaturtlesandwich:
found a workaround for now by using winres to set this manifest
```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1" xmlns:asmv3="urn:schemas-microsoft-com:asm.v3">
    <asmv3:application>
        <asmv3:windowsSettings>
            <dpiAware xmlns="http://schemas.microsoft.com/SMI/2005/WindowsSettings">true</dpiAware>
            <dpiAwareness xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">PerMonitorV2</dpiAwareness>
        </asmv3:windowsSettings>
    </asmv3:application>
</assembly>
```

bananaturtlesandwich:
Fixed with my pr #327

PR #327, opened 2022-11-26
fix for #326

bananaturtlesandwich:
# Changes
- I don't think there was any need to load the libraries when the function was already in winapi so I removed the lib loading
- I think calling `SetProcessDpiAwareness` after making fullscreen was causing #326 so I moved the call to before
- I changed `SetProcessDpiAwareness` to `SetProcessDPIAware` since I'm sure it's dpi unaware anyway no point to reassign
- I tried to leave as much of the other internals intact (PLEASE TELL ME IF I BROKE SOMETHING)
# Difference
- Now if I try to borderless fullscreen and high_dpi at the same time #326 doesn't happen
- setting high_dpi by itself also doesn't hide the title and icon (I forgot to mention that issue in #326)

### DIFF
--- a/src/native/windows.rs
+++ b/src/native/windows.rs
@@ -8,13 +8,13 @@ use crate::{
 use winapi::{
     shared::{
         hidusage::{HID_USAGE_GENERIC_MOUSE, HID_USAGE_PAGE_GENERIC},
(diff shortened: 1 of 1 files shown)


not-fl3:
Wow, less code is always better, thanks for PR!

PR #331, opened 2022-12-14
gl.js: use performance timer

troyc:
Use performance.now instead of Date.now, which may provide a higher-resolution time (depending on the browser's Spectre mitigation settings). The time returned will no longer be relative to the Unix epoch and system clock.


troyc:
I amended the commit to add performance.timeOrigin to the returned result, which should be equivalent to Date.now() with the added benefit of sub-millisecond timing for browsers that support it. I explored this possibility with the following jsfiddle:

```
let date = Date.now()
let now = performance.now()
//performance.timing.navigationStart is deprecated
let now2 = performance.timing.navigationStart + performance.now()
let now3 = performance.timeOrigin + performance.now()

console.log("date: " + date)
console.log("now: " + now)
console.log("now2: " + now2)
console.log("now3: " + now3)
```

PR #332, opened 2022-12-29
Fixed panic on paste in WASM

### LINKED ISSUE #304: Panic on paste for WASM (with egui-miniquad)
I made an app with egui-miniquad. It works fine locally, but when using wasm pressing Ctrl+V causes a panic with the following error: 
already borrowed: BorrowMutError
in miniquad-0.3.8\\src\\native\\wasm.rs", line: 95, col: 35

What seems to be happening is:
`miniquad src/native/wasm.rs line 343` has `key_down()`, which uses `with(|globals|{})`, doing the first borrow.
This calls my event handler, which as recommended calls `egui_mq.key_down_event()`
When doing a paste, `egui-miniquad src/lib.rs line 291` calls `get_clipboard()` which in turn calls `mq_ctx.clipboard_get()` from miniquad
`miniquad src/native/wasm.rs line 63` has `clipboard_get()` which calls a global `clipboard_get()` from `line 259` which does another call to `with(|globals|{})`. This tries to do a second borrow, which fails.

I think this could be fixed by replacing the `src/native/wasm.rs line 63` call with a clone of the clipboard value which is accessible due to the reference passed into the function. If you'd like I could test this fix and then submit a pull request.

There also seems to be an issue where ctrl+C needs to be used twice for it to actually be copied into the OS clipboard. I'm not quite sure what's going on here, but I thought I'd mention it since I noticed it while working on this issue.

mkhan45:
Fixes #304. Thanks @ssk97!

### DIFF
--- a/src/native/wasm.rs
+++ b/src/native/wasm.rs
@@ -61,7 +61,7 @@ impl NativeDisplay for WasmDisplay {
         }
     }
     fn clipboard_get(&mut self) -> Option<String> {
-        clipboard_get()
+        self.clipboard.clone()
     }
     fn clipboard_set(&mut self, data: &str) {
         clipboard_set(data)


not-fl3:
oh, good catch, thanks!

PR #337, opened 2023-01-28
Add GL_TRIANGLE_STRIP and GL_LINE_STRIP to PrimitiveType Enum

tanolino:
Add missing OpenGL primitive types

Explaining what they do:
https://learnopengl.com/Advanced-OpenGL/Geometry-Shader

They appear to be defined since OpenGL 2
https://docs.gl/gl2/glDrawElements
so I think they are available on all supported platforms.

### DIFF
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -1407,14 +1407,18 @@ impl From<CompareFunc> for GLenum {
 #[derive(Debug, PartialEq, Clone, Copy)]
 pub enum PrimitiveType {
     Triangles,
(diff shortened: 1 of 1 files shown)


tanolino:
The reason why I would like to have these options is that I am puzzling together a 2D RPG-Maker - similar engine (nothing too serious) and GL_TRIANGLE_STRIP would help to only use 4 instead of 6 Vertex per rectangle.

not-fl3:
Thanks!

PR #338, opened 2023-01-31
0.4

not-fl3:


src/graphics.rs
```diff
@@ -2004,3 +979,192 @@ impl ElapsedQuery {
...
 }
+
+/// A vtable-erased generic argument.
+/// Basically, the same thing as fn f<U>(a: &U), but
+/// trait-object friendly.
+pub struct Arg<'a> {
+    ptr: *const std::ffi::c_void,
+    element_size: usize,
+    size: usize,
+    is_slice: bool,
```
maciejhirsz, line 990: `is_slice` isn't being used for anything that I can see except for an assert. Could maybe change those to `debug_assert!` and put the flag behind `#[cfg(debug_assertions)]`?

not-fl3: hey I am actually very intrested in you opinion here, I am super unhappy with the whole Arg thing 

`is_slice` is used to protect against one particular bug: 

This is the correct usage of `new_buffer_*`
```
let buffer = [1, 2, 3];
new_buffer_stream(.., Arg::slice(&buffer));
```

but this will also compiles well... 
```
let buffer = [1, 2, 3];
new_buffer_stream(.., Arg::val(&buffer));
```

I do not want to merge it like this

One of the solutions is to have two distinct types, something like `new_buffer_stream(.., SlicePtr::new(&slice))` and `apply_uniforms(.., Ptr::new(&uniforms))`

but I just really, really dislike the naming :|

not-fl3: Another solution that looks beautiful, but is horrible on the implementation side is 

```
trait RenderBackend {
    fn new_buffer_from_bytes(&self, ptr: *mut (), each_size: usize, count: usize);
}

trait NotObjectSafeTraitWithNiceNames {
    fn new_buffer_stream<T>(&self, buffer: &[T]);
}

impl<T: RenderBackend + ?Sized> NotObjectSafeTraitWithNiceNames for T {
    fn new_buffer_stream<T1>(&self, buffer: &[T1]) {
        self.new_buffer_from_bytes(
            buffer.as_ptr() as _,
            std::mem::size_of::<T1>(),
            buffer.len(),
        )
    }
}

struct GlContext;
impl RenderBackend for GlContext {
    fn new_buffer_from_bytes(&self, ptr: *mut (), each_size: usize, count: usize) {
        println!("{} {}", each_size, count);
    }
}

fn main() {
    let foo: Box<dyn RenderBackend> = Box::new(GlContext);

    // this looks nice!
    foo.new_buffer_stream(&[1, 2, 3]);

  // but will require importing the second trait... otherwise, if we the trait is in the module/crate:
  // help: the following trait is implemented but not in scope; perhaps add a `use` for it:
  // |
  // 1  | [use miniquad::NotObjectSafeTraitWithNiceNames;]
  // |
}
```

not-fl3: also, good point, changed `assert` to `debug_assert`, thanks!

maciejhirsz: I thought you wanted to use the `Arg` to avoid codegen bloat, though I'm not sure it necessarily helps.

You could just have a helper trait implemented for all slices:

```rust
trait AsArg {
    fn as_arg(&self) -> Arg;
}

impl<T> AsArg for [T] {
    fn as_arg(&self) -> Arg {
        Arg::slice(self)
    }
}
```

and then just convert it at the beginning of each method that needs it:

```rust
fn new_buffer_immutable(&mut self, buffer_type: BufferType, data: impl AsArg) -> BufferId {
    let data = data.as_arg();
    // ...
}
```

If you want to really avoid codebloat you can use dyn dispatch here too the trait is implemented on the unsized `[T]`:

```rust
fn new_buffer_immutable(&mut self, buffer_type: BufferType, data: &dyn AsArg) -> BufferId {
    let data = data.as_arg();
    // ...
}
```

`Arg` here is a pretty high-gravity term, I'd probably change it to something more verbose and specific, like `GlBuffer` or some such (and rename the trait appropriately).

For `apply_uniforms` I don't know that there is really much of a point in guarding anything by some code magic, but you could do the same thing using a different trait that's implemented for all `T`. That said, if you really want to make it correct by construction the Rust way about it would be a proc macro, so instead of:

```rust
pub fn meta() -> ShaderMeta {
    ShaderMeta {
        images: vec![],
        uniforms: UniformBlockLayout {
            uniforms: vec![UniformDesc::new("mvp", UniformType::Mat4)],
        },
    }
}

#[repr(C)]
pub struct Uniforms {
    pub mvp: glam::Mat4,
}
```

You could do:

```rust
pub fn meta() -> ShaderMeta {
    ShaderMeta {
        images: vec![],
        uniforms: Uniforms::layout(), // provided by GlUniforms trait
    }
}

#[derive(GlUniforms)]
#[repr(C)]
pub struct Uniforms {
    pub mvp: glam::Mat4,
}
```

Obviously pulling in a derive proc macro will increase compile times so there is a cost here. You could amortize it in macroquad by manually deriving everything though.

BTW: since uniforms are unchanging after compile time (correct me if I'm wrong here), the layout could be described using  `&'static [_]` slices instead of `Vec<_>`.

not-fl3: > fn new_buffer_immutable(&mut self, buffer_type: BufferType, data: impl AsArg) -> BufferId {

I would be supper happy with `data: &[T]`, but I can't really use neither impl trait or generic in a trait function :(

not-fl3: All this mess is because before we had just GlContext and it had 
`fn some<T>(&self, buffer: &[T])` and it was perfect! 

but now, with GlContext/MetalContext, I want to have a trait and in a trait I can't really receive a pointer + a bit of a type info :(

maciejhirsz: Ah, gotcha, you can parameterize it on the trait itself:

```
pub trait RenderingBackend<T> {
    fn new_buffer_immutable(&mut self, buffer_type: BufferType, data: &[T]) -> BufferId;
}
```

not-fl3: Called it `BufferSource::slice` and `UniformsSource::table`. Actually, I kind of like it now!

The final PR with metal: https://github.com/not-fl3/miniquad/pull/344 :)

PR #341, opened 2023-02-04
add GL_POINTS

bananaturtlesandwich:
4a2fe20e1145855318b2f744bd00c263e4022ed6 added triangle strip and line strip but forgot about points

### DIFF
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -1410,6 +1410,7 @@ pub enum PrimitiveType {
     TriangleStrip,
     Lines,
     LineStrip,
+    Points,
 }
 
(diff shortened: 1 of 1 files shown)


not-fl3:
I can't find a link, but I do remember that GL_POINTS (and, maybe, GL_LINES) support was very limited on some gpus.. not really sure if adding them is really worth the effort on supporting/software emulating potential problems

(I can find a lot of links to broken gl_PointSize, but, iirc, there was more than just PointSize)

bananaturtlesandwich:
Interesting I could've sworn it was supported on all platforms

PR #348, opened 2023-03-10
center window on create and set_window_size()

RehkitzDev:
centers the window on the main screen.
changes only affect windows platform.

probably nice to have on linux & macos too.

### DIFF
--- a/src/native/windows.rs
+++ b/src/native/windows.rs
@@ -110,17 +110,10 @@ impl crate::native::NativeDisplay for Display {
         self.user_cursor = cursor_icon != CursorIcon::Default;
     }
     fn set_window_size(&mut self, new_width: u32, new_height: u32) {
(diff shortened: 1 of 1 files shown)


not-fl3:
To be honest I would prefer window manager to do its job and place window accordingly, at least by default. 

Maybe add a flag in `conf.platform` to make this behavior optional?

RehkitzDev:
Where is the window manager you mentioned? Add set_window_position in NativeDisplay? Currently only have the ability to add this on windows.
I'm not sure about adding a flag for this. Who would want the default windows behavior? It spawns the window almost at the top left of the screen.

not-fl3:
I just believe that apps should not manage their window placement, its the job for window manager (or the entire os in case of windows/mac), so I would prefer anything forcing window size/position to be optional :)

PR #349, opened 2023-03-10
Allow setting paired texture parameters separately

jabuwu:
This PR adds two functions to textures: `set_filter_min_mag` and `set_wrap_xy`

The existing functions, `set_filter` and `set_wrap`, are probably fine for 99% of use cases, so I think it's better to add these new specialized functions rather than changing the existing ones, but we can change this however it will work best.


not-fl3:
Thanks for PR!

PR #351, opened 2023-03-14
resolve #165

### LINKED ISSUE #165: sapp-windows: Implement ABSOLUTE raw device mouse events
Windows do not send MOUSE_MOVE_ABSOLUTE on WM_INPUT event, but apparently, it can. 
Would be nice to support MOUSE_MOVE_ABSOLUTE. 

Related comment: https://github.com/not-fl3/miniquad/pull/164/files#r550873182
The code: https://github.com/not-fl3/miniquad/blob/763cc21f41467b2af869d9d29c12d5c2a2d5e716/native/sapp-windows/src/lib.rs#L766

bananaturtlesandwich:
implements MOUSE_MOVE_ABSOLUTE so #165 is finally resolved

> yeah, I'm not sure. glfw has a bunch of code to track the last known position and still provide deltas even if the hardware uses absolute coordinates, but I heard hardware very rarely if ever gives off these events. sure enough, I tested all of the mice I had and they all gave off relative mouse coords. I think support for absolute coords should only be added if someone is actually running into this panic, because I think ABSOLUTE is just legacy at this point.

well someone using an app I made in miniquad is encountering the panic
I'm pretty sure all that was missing was conversion from normalised absolute coordinates so I've implemented that
Also simplified the high dpi assignment I wrote in #327

### DIFF
--- a/src/native/windows.rs
+++ b/src/native/windows.rs
@@ -444,12 +444,15 @@ unsafe extern "system" fn win32_wndproc(
                 panic!("failed to retrieve raw input data");
             }
 
-            if data.data.mouse().usFlags & MOUSE_MOVE_ABSOLUTE == 1 {
-                unimplemented!("Got MOUSE_MOVE_ABSOLUTE on WM_INPUT, related issue: https://github.com/not-fl3/miniquad/issues/165");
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks for PR!

PR #354, opened 2023-03-19
Support for keys Shift, Control, Alt and Cmd on MacOS.

IohannRabeson:
Hi! On MacOS, it's a little bit harder to know if the keys Shift, Control, Alt and Cmd are pressed or not.


not-fl3:
Awesome, thanks a lot for PR!

PR #361, opened 2023-04-07
fix: Display initialized too late for fullscreen

Bright-Shard:
On v0.4 on macOS, the tl_display's display is initialized too late for fullscreen to work (you can see this with the window_conf example). This just bumps window initialization up a few lines so that's no longer an issue.


not-fl3:
Thanks!

PR #362, opened 2023-04-07
fix: Incorrect Windows conditional compilation

Coca162:
This fixes the conditional compilation on v0.4 to make Window's events/fullscreen work again


not-fl3:
Thanks!

PR #363, opened 2023-04-09
fix: Touch events on iOS (v0.4)

Bright-Shard:
Currently in v0.4, iOS taps translate into mouse click events in the event handler. This fix will instead call the touch event, and passes the ID for each touch. I tested this with the blobs example.


not-fl3:
Thanks!

PR #365, opened 2023-04-14
Fixes for x11 and focus events for x11 and wasm

jusax23:
Added the posebility to leave fullscreen on x11.
- Not 100% sure this is the correct way, but better then befor.

Implemented set_window_size for x11.

Fixed Default Cursor for x11.
- Befor the change the cursor was set to XC_left_ptr, but this is not always the default cursor. (e.g. with changed System default)
- Fix was to use XUndefineCursor to withdraw cursor changes.

Added calling window_minimized_event and window_restored_event for x11 and wasm triggered on focus change events.
- The Name of this events are maybe not the best but the calls on android could also be seen as focus change events.
- I thing implementing it this way is better then adding a new event.

### DIFF
--- a/js/gl.js
+++ b/js/gl.js
@@ -1253,6 +1253,18 @@ var importObject = {
                 wasm_exports.on_files_dropped_finish();
             };
 
+            let lastFocus = document.hasFocus();
+            var checkFocus = function() {
+                let hasFocus = document.hasFocus();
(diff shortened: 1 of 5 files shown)


not-fl3:
Hi! Could you please split it into multiple PRs with one feature per one PR?

jusax23:
Okay! I will do this and close this PR.

PR #366, opened 2023-04-14
Adding leave fullscreen for x11.

jusax23:
- Changed XEvent and XPropertyChange when fullscreen var is false.
- Not 100% sure this is the right way, but seems to work better then befor. 
Remaining TODO:
- Restore window size after fullscreen leave.


not-fl3:
Thanks for PR!

I am not sure how it is supposed to work either, but if it works - it is for sure better than before, thanks again :)

PR #368, opened 2023-04-14
Fixing default Cursor for x11

jusax23:
When the default System cursor was changed XC_left_ptr is not the default Cursor.
- Added XUndefineCursor to libx11.rs
- Changed to calling XUndefineCursor, when calling set_cursor with CursorIcon::Default

### DIFF
--- a/src/native/linux_x11.rs
+++ b/src/native/linux_x11.rs
@@ -634,32 +634,38 @@ impl X11Display {
         let libx11 = &mut self.libx11;
         let display = self.display;
 
-        let cursor = match cursor {
-            None => {
-                // empty_cursor was created during create_window
-                self.empty_cursor.unwrap()
-            }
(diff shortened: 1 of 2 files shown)


src/native/linux_x11.rs
```diff
@@ -634,32 +634,38 @@ impl X11Display {
...
-                        CursorIcon::EWResize => libx11::XC_sb_h_double_arrow,
-                        CursorIcon::NSResize => libx11::XC_sb_v_double_arrow,
-                        CursorIcon::NESWResize => libx11::XC_top_right_corner,
-                        CursorIcon::NWSEResize => libx11::XC_top_left_corner,
-                    },
-                )
-            }),
-        };
-        (libx11.XDefineCursor)(display, window, cursor);
+        if cursor == Some(CursorIcon::Default) {
```
not-fl3, line 637: I understand the problem, but I do not really like the extra branching and a special case for CursorIcon::Default, can't really propose anything better right now tho, will take a closer look into it a bit later

PR #369, opened 2023-04-14
Window focus events

jusax23:
Added window_minimized_event and window_restored_event for linux_x11 and wasm.
These events will be called on focus change events.
The names of this events are maybe not the best,
but I think this way is better then implementing a new event?!
changes:
- added extern fn focus in wasm.rs
- added a call to this fn in gl.js
- added calls from XEvent 9 and 10 to the window_events


not-fl3:
*sorry for delay with reviews* 

Thanks!

PR #371, opened 2023-04-21
added get_gl_proc_addr to GraphicsContext

limikael:
It would be nice to be able to use miniquad together with [gl](https://crates.io/crates/gl), as is now possible with SDL, Glutin and glfw. Currently it is not possible with miniquad, since miniquad doesn't expose any way to get an OpenGL proc address.

This PR adds this functionality to GraphicsContext with a new function `get_gl_proc_addr`.

It is currently only implemented for X11. Adding this PR to see is the idea as well and the way it is currently implemented for X11 is well received.

### DIFF
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -1,4 +1,4 @@
-use std::{ffi::CString, mem};
+use std::{ffi::{CString, c_void}, mem};
 
 mod texture;
 
@@ -653,6 +653,8 @@ pub struct GraphicsContext {
 
     pub(crate) features: Features,
     pub(crate) display: Option<*mut dyn crate::NativeDisplay>,
+
+    pub(crate) gl_proc_addr_getter: Option<Box<dyn Fn(&str)->*const c_void>>
 }
 
 impl GraphicsContext {
@@ -668,6 +670,7 @@ impl GraphicsContext {
             glGenVertexArrays(1, &mut vao as *mut _);
             glBindVertexArray(vao);
             GraphicsContext {
+                gl_proc_addr_getter: None::<Box::<dyn Fn(&str)->*const c_void>>,
                 default_framebuffer,
                 shaders: vec![],
                 pipelines: vec![],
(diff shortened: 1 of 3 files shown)


limikael:
Oh btw, the way to use it (in the context of [this example](https://github.com/not-fl3/miniquad/blob/150426de6e856aa9f87077709e0f6fd075679f2c/examples/quad.rs#L88)), would be:

```rust
fn main() {
        gl::load_with(|s|ctx.get_gl_proc_addr(s)); // Gl bindings for the gl crate will now be available.
        Box::new(Stage::new(&mut ctx))
    });
}
```

not-fl3:
I am not sure if this is in scope of miniquad, considering lack of glGetProcAddr on web, miniquad's own gl/metal loader. If there is a requirement to use `gl` crate with miniquad - maybe consider a fork?

limikael:
Hm ok... Just to ask for clarification how it is outside the scope...

Is it because miniquad is intended to be a "graphics library", not a window/context creation library? And the window and context creation is just there to support the graphics and drawing?

The fact is the the window and context creation in miniquad is very good, one of the best and slickest out there. So it would be great to be able to use just that part of miniquad. To provide some background on what I'm trying to do, I created this repo:

https://github.com/limikael/glapp

So it is a kind of meta library, that wraps different other libraries. Would love to have miniquad there as well, and it would be kind of sad to have to fork it, because that would make maintenance more difficult in the future. An for me, it is very valuable to get the exposed get_proc_address just on the platforms that support it, even if it is not available on all platforms.

not-fl3:
yes, I do believe that implementing this properly, for more platforms etc will require significant changes to miniquad, that I personally would try to avoid doing/maintaining

notgull:
You could maybe consider returning a [`glow`] `Context` instead, since that bridges the web/desktop gap for OpenGL.

limikael:
> You could maybe consider returning a [`glow`](https://crates.io/crates/glow) `Context` instead, since that bridges the web/desktop gap for OpenGL.

This would require similar changes to miniquad, which is something the author does not want. There is now a fork, as the author recommended, with the goal of just using the window and context creation from miniquad, and expose a `get_procaddr` function for GL. https://github.com/loki-chat/lokinit

PR #376, opened 2023-06-06
Fix offset in glDrawElementsInstanced

mario-holubar:
Noticed this trying to get imgui to work


not-fl3:
oh, good catch, thanks!

not-fl3:
it was a big mess with the index_type, the comment claimed it is 1, 2, 4, but in fact it stored some GL_* constants..

Fixed with https://github.com/not-fl3/miniquad/commit/ef68e06ac6c953671cf3be6c4ab4c748e919490c, hope it still works!

PR #378, opened 2023-06-14
Clear cache when deleting buffer

### LINKED ISSUE #321: Stream buffer recreating segfaults
This called in the draw method causes a segfault on Windows and MacOS. ( (exit code: 0xc0000005, STATUS_ACCESS_VIOLATION))
```
let indices = [0u16, 1, 2, 0, 2, 3];
self.bindings.index_buffer.delete(); // without delete it "works"
self.bindings.index_buffer = Buffer::stream(ctx, BufferType::IndexBuffer, indices.len() * std::mem::size_of::<u16>());
self.bindings.index_buffer.update(ctx, &indices);
```

Full example:
https://gist.github.com/pum-purum-pum-pum/c399452086016fbb3f5ff69aa4f8a4e6

Windows machine I've tried it on:

```
GPU: NVIDIA quadro p5000
Driver provider: NVIDIA
```
Also reproduced on M1 laptop and M1 mac mini.

Original issue https://github.com/not-fl3/egui-miniquad/issues/42

mario-holubar:
Fixes #321

### DIFF
--- a/src/graphics/gl.rs
+++ b/src/graphics/gl.rs
@@ -1044,6 +1044,8 @@ impl RenderingBackend for GlContext {
     /// this function is not marked as unsafe
     fn delete_buffer(&mut self, buffer: BufferId) {
         unsafe { glDeleteBuffers(1, &self.buffers[buffer.0].gl_buf as *const _) }
+        self.cache.clear_buffer_bindings();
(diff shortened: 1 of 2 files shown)


not-fl3:
Thanks for PR!

PR #379, opened 2023-06-14
Allow reallocation of gl buffer

mario-holubar:
When buffer_update() is called with a size greater than the buffer's original size, reallocate it using glBufferData()

Or is the lack of this intentional to incentivize keeping buffers at a fixed size?


mario-holubar:
Ah yeah that makes sense. I've never used Metal so I don't know if it's possible.

PR #380, opened 2023-06-15
Remove unused vertex array

mario-holubar:


not-fl3:
To be honest I can't give a good explanation right away (it should be in the comment above this code!), but I am not sure if not having any active binded VAO will not break anything.

sokorototo:
This PR breaks rendering in general, the VAO is actually important

PR #384, opened 2023-07-09
fix(windows): fix endless resize events while minimized

Corbanator:
When a window is minimized, the dimensions of the client rect are set to (0,0), while the dimensions in the display data are set to a minimum of (1,1), causing them to always differ. This causes it to endlessly send resize events when minimized. Moving the `max(1)` calls to before the comparison fixes this issue.


not-fl3:
Good catch, thanks for PR!

Issue #343, opened 2023-02-11
Disable window decorations on wayland


narodnik:
I initialize the library with:

```rust
struct Stage {
}

impl Stage {
    fn new(ctx: &mut miniquad::Context) -> Self {
        Self {}
    }
}

impl miniquad::EventHandler for Stage {
    fn update(&mut self, _ctx: &mut miniquad::Context) {}

    fn draw(&mut self, ctx: &mut miniquad::Context) {
        ctx.clear(Some((0., 1., 0., 1.)), None, None);
    }
}

fn main() {
    miniquad::start(miniquad::conf::Conf {
        window_resizable: true,
        platform: miniquad::conf::Platform {
            linux_backend: miniquad::conf::LinuxBackend::WaylandOnly,
            ..Default::default()
        },
        ..Default::default()
    }, |mut ctx| {
        Box::new(Stage::new(&mut ctx))
    });
}

```

I get this message: "Decoration manager not found, will draw fallback decorations"

Then miniquad draws a titlebar and window borders. I saw it is doing this in the code. Is there any way to disable all window decorations?

not-fl3:
Not really :( 

we probably should add some configuration into `conf::platform` for this

narodnik:
In `src/native/linux_wayland.rs`
```rust
        if display.decoration_manager.is_null() {
            eprintln!("Decoration manager not found, will draw fallback decorations");
        }

        // ...

        if display.decoration_manager.is_null() == false {
            // ...
        } else {
            //display.decorations = Some(decorations::Decorations::new(
            //    &mut display,
            //    conf.window_width,
            //    conf.window_height,
            //));
        }
```
Commenting the else branch works fine. Is there any reason why it's forcing window decorations? Can I delete this code? Or if you prefer it as a config option, should I add it to the end of `conf::Platform` as `wayland_no_force_decoration`?

PR #385, opened 2023-07-16
wayland: disable drawing decorations when decoration manager is not found

narodnik:
See issue #343 
I run a tiling window manager, and miniquad attempts to draw a window border and titlebar. This behaviour is generally undesirable, and unneeded.


not-fl3:
I think this should be under some sort of a configuration flag - on a non-tiled window managers (see gnome) no-decorations windows looks really awful :(

narodnik:
Closed this pull request and opened #386 instead to make it configurable via a flag.

PR #386, opened 2023-07-17
wayland: add a flag to disable drawing decorations when decoration manager is not found

narodnik:
See issue #343 
I run a tiling window manager, and miniquad attempts to draw a window border and titlebar. This behaviour is generally undesirable, and unneeded.

Previously pull request #385 deleted the offending code but I've instead made it configurable via a flag as discussed in the pull request.


not-fl3:
Perfect, thanks for PR!

PR #388, opened 2023-07-19
Add missing wayland keyboard input

narodnik:
Add missing keyboard input support. The only thing I didn't like was using the static global EVENTS to communicate with the
handlers. Unfortunately, I'm not able to store the event handler in display, and cannot pass it as data since it happens where the seat is attached.

You can test this code by modifiying `examples/triangle.rs` like so
```diff
diff --git a/examples/triangle.rs b/examples/triangle.rs
index a1b1104..0b4e6a5 100644
--- a/examples/triangle.rs
+++ b/examples/triangle.rs
@@ -82,9 +82,14 @@ impl EventHandler for Stage {
 
         self.ctx.commit_frame();
     }
+
+    fn key_down_event(&mut self, keycode: KeyCode, modifiers: KeyMods, repeat: bool) {
+        println!("hello {:?} {:?} {}", keycode, modifiers, repeat);
+    }
 }
 
 fn main() {
+    /*
     let mut conf = conf::Conf::default();
     let metal = std::env::args().nth(1).as_deref() == Some("metal");
     conf.platform.apple_gfx_api = if metal {
@@ -94,6 +99,19 @@ fn main() {
     };
 
     miniquad::start(conf, move || Box::new(Stage::new()));
+    */
+    miniquad::start(
+        miniquad::conf::Conf {
+            window_resizable: true,
+            platform: miniquad::conf::Platform {
+                linux_backend: miniquad::conf::LinuxBackend::WaylandOnly,
+                wayland_use_fallback_decorations: false,
+                ..Default::default()
+            },
+            ..Default::default()
+        },
+        || Box::new(Stage::new()),
+    );
 }
 
 mod shader {
```


not-fl3:
Great work, thanks! 

For the statics - well, thats not ideal, but I also do not see the way to do it better (see SEAT_HANDLER that was a static already).

PR #398, opened 2023-07-30
return the texture_set_wrap_xy and texture_set_filter_min_mag methods

lenscas:
I am not sure how to test this, so all I did was make sure it compiles


lenscas:
I can't test metal, but this branch does seem to work with my project so I think the "not sure how to test it" statement from before is wrong.

not-fl3:
Thanks for PR, sorry for merging delay :)

lenscas:
np, it happens. Thanks for merging :)

PR #400, opened 2023-08-05
wayland treats shifted latin chars as a different keysym so handle that.

narodnik:
FYI shift is already handled in the main event loop where we apply other modifiers like ctrl, alt, .etc


not-fl3:
Thanks a lot for all your work keeping miniquad/wayland afloat!

PR #409, opened 2023-09-20
Misc fixes to get miniquad programs run on KDE Wayland

hch12907:
A demonstration of server-side decoration:

image

### DIFF
--- a/src/native/linux_wayland.rs
+++ b/src/native/linux_wayland.rs
@@ -678,6 +678,15 @@ where
             &mut display as *mut _ as _,
         );
 
+        let title = std::ffi::CString::new(conf.window_title.as_str()).unwrap();
+
+        wl_request!(
+            display.client,
+            display.xdg_toplevel,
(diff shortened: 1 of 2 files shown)


src/native/linux_wayland.rs
```diff
@@ -397,7 +397,7 @@ unsafe extern "C" fn registry_add_object(
                 1,
             ) as _;
         }
-        "zxdg_decoration_manager" => {
+        "zxdg_decoration_manager_v1" => {
```
not-fl3, line 400: We should probably keep both, zxdg_decoration_manager and zxdg_decoration_manager_v1?

hch12907: Hmm, both compositors I have (sway & kwin) use zxdg_decoration_manager_v1.

hch12907: Done

not-fl3:
Thanks for PR!

Issue #299, opened 2022-06-21
MSAA doesn't work on linux


Mivik:
The `desired_sample_count` used [here](https://github.com/not-fl3/miniquad/blob/master/src/native/linux_x11/glx.rs#L406) is always `0` and thus the MSAA is always disabled `linux_x11`.

Also, I noticed that `Conf::sample_count` is not even used in `linux_wayland`, indicating that the both native implementations have problems with multisampling.

PR #413, opened 2023-11-14
fix MSAA doesn't work on linux

Lishen1:
https://github.com/not-fl3/miniquad/issues/299#issue-1277885151


not-fl3:
Thanks for PR!

PR #415, opened 2024-01-27
Fixed forced modal pause of the application when resizing or moving on Windows

Shelim:
# The problem

Current implementation will force a pause on resizing or moving a window on Windows. This pause is modal: it will block all activity in main thread, including all polling inside update (which can result in as severe consequences as disconnection of multiplayer games!).

prev

# The solution

The solution was long-and-length discussed in [8-years-spanning thread](https://github.com/libsdl-org/SDL/issues/1059) at SDL repo (!) and finally met final, safe solution in this [SDL 2.30 PR](https://github.com/libsdl-org/SDL/commit/509c70c6982b6927f5a8d4fb32f9319cbaf0c2ef) at 2023, November.

This is a port of that solution into miniquad. It will only affect Windows platform.

fixed

### DIFF
--- a/src/native/windows.rs
+++ b/src/native/windows.rs
@@ -48,6 +48,7 @@ pub(crate) struct WindowsDisplay {
     wnd: HWND,
     dc: HDC,
     event_handler: Option<Box<dyn EventHandler>>,
+    modal_resizing_timer: usize,
 }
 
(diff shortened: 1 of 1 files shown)


not-fl3:
Great work, thanks!

PR #420, opened 2024-03-07
fix scrolling on windows

DissolveDZ:
scrolling on windows would return a value 120 times bigger than it should be rather than 1
this pull request just uses .signum() to clamp the range that mouse_wheel().1 would return

### DIFF
--- a/src/native/windows.rs
+++ b/src/native/windows.rs
@@ -433,11 +433,11 @@ unsafe extern "system" fn win32_wndproc(
             // );
         }
         WM_MOUSEWHEEL => {
-            event_handler.mouse_wheel_event(0.0, (HIWORD(wparam as _) as i16) as f32);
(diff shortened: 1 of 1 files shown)


DissolveDZ:
I'm not quite sure what the OS values would represent as i never changed much and i was kind of thrown off by it after porting my game to windows

DissolveDZ:
if found the following in the official documentation
`The wheel rotation will be a multiple of WHEEL_DELTA, which is set at 120. This is the threshold for action to be taken, and one such action (for example, scrolling one increment) should occur for each delta.`

DissolveDZ:
so i'm guessing the mouse itself controls this mouse delta value but this would result in inconsistent scrolling across different mouses so you could make this some kind of function like mouse_scroll_delta or smth. like that due to most people not knowing that this is a threshold and not the actual scroll and that it could differ from mouse to mouse

PR #421, opened 2024-03-12
Add clear_background option to window_conf

DuckyBlender:
Needed for https://github.com/not-fl3/macroquad/pull/705
Unfortunately doesn't work yet


not-fl3:
Just to keep the discussion in one place, lets continue this here: https://github.com/not-fl3/macroquad/pull/705#issuecomment-1995981761

PR #424, opened 2024-04-10
Fix dpi_scale() always returning 1 on Windows

InZan17:
dpi_scale() would always return 1 on Windows despite setting high_dpi to true and my dpi settings at 150%.
This is fixed by calling display.init_dpi before setting dpi_scale to display.window_scale.


not-fl3:
Thanks for PR!

PR #425, opened 2024-04-14
Fix resizing managed textures

vdrn:
Resizing texture does not update its size params (since `self.textures.get(..)` returns a `Copy`)


not-fl3:
Thanks for PR!

Issue #329, opened 2022-12-10
feature request: API to get/set window position


trevyn:
Like `set_window_size`, but allow getting and setting the window's position on the desktop.

PR #434, opened 2024-05-15
Add {get/set}_window_position for Windows platform.

mandroll:
Add `get_window_position` and `set_window_position` to miniquad for Windows platform.

- `set_window_position` uses the typical Request code path.
- `get_window_position` adds a `screen_position` field to `NativeDisplay` struct and updates it every time the screen position changes or is resized in the main update loop.

## Background

My motivation for this is to implement some form of pseudo hot-reloading (recompiling and reloading game state), which requires the window to always be in the same position and size. I've included an example in this [gist](https://gist.github.com/mandroll/b2a0a0dea831c601b67deb56abbdb38f), with a short video demonstration [here](https://youtu.be/n42oIPoRCLU).

There have also been numerous requests for this feature in the Discord channel and also #329.

## Other platforms

This PR only adds Windows support. Other OSes will give an unimplemented error when trying to use these features. It is not applicable to ios/android/wasm platforms. For other platforms (`linux_*`, `macos`), adding support involves:

1. Add a `set_window_position` implementation and calling it in `process_request`
2. Update `NativeDisplay.screen_position` in the main update loop for that platform.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -218,6 +218,8 @@ pub mod window {
     }
 
     /// Get the position of the window.
+    /// TODO: implement for other platforms
+    #[cfg(target_os = "windows")]
     pub fn get_window_position() -> (u32, u32) {
         let mut d = native_display().lock().unwrap();
         d.screen_position
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -212,6 +212,17 @@ pub mod window {
         });
     }
 
+    pub fn set_window_position(new_x: u32, new_y: u32) {
+        let mut d = native_display().lock().unwrap();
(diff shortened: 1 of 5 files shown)


examples/save_window.rs
```diff
@@ -0,0 +1,75 @@
+/// Example of saving window position to enable "hot reloading".
```
not-fl3, line 1: its a very niche and super platform-specific example, maybe keep it in some repo outside of miniquad itself? there might be a separate examples repo linked from `awesome-quads` list or something

mandroll: That's fair. Removed.

not-fl3:
Thanks for PR!

PR #438, opened 2024-05-26
Add support for PrimitiveType::Points

rlofc:
This will allow using PrimitiveType::Points when creating a new Material in macroquad, like so:
```rust
let material = load_material(
    ShaderSource::Glsl {
        vertex: VERTEX_SHADER,
        fragment: FRAGMENT_SHADER,
    },
    MaterialParams {
        pipeline_params: PipelineParams {
            // Added PrimitiveType::Points can be used in PipelineParams to render points rather than
            // lines or triangles.
            primitive_type: miniquad::PrimitiveType::Points,
            color_blend: Some(blend_mode),
            alpha_blend: Some(BlendState::new(
                miniquad::Equation::Add,
                miniquad::BlendFactor::Zero,
                miniquad::BlendFactor::One,
            )),
            ..Default::default()
        },
        ..Default::default()
    },
)
```
and is useful for various particle effects.

### DIFF
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -646,6 +646,7 @@ impl Default for Equation {
 pub enum PrimitiveType {
     Triangles,
     Lines,
+    Points,
 }
 
 impl From<PrimitiveType> for GLenum {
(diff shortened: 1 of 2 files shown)


not-fl3:
Thanks for PR!

PR #440, opened 2024-05-30
add explicit keycode values based off X11's keysymdef.h

narodnik:
This is useful for serialization of KeyCode values. I used keysymdef.h from X11 as the standard values. We use u16 but this can be changed to u32 if needed in the future. For values missing in the list, I used a prefix of 0x01.

The missing keycode values I couldn't find in keysymdef.h were: World1, World2 and Unknown. Unknown doesn't exist for obvious reasons, but I'm not sure what World1 and World2 are meant to be. From keysymdef.h, and grepping the code, seems World1 corresponds to `#define XK_Select                        0xff60  /* Select, mark */`? And world2 is only returned for the windows keycodes?


not-fl3:
I am not sure those values should be connected to X11 once, I feel like 
```
pub enum KeyCode {
    Space = 0,
```
would work just as well?

narodnik:
I think it's good to have a standard, and the values make sense. We don't have to be strict about it, but I originally tried using ASCII codes for the values, but then there's a bunch of keys like Insert that aren't ASCII. In the future we may add more exotic keys so `u16` seemed like a decent enough compromise.

But if you prefer starting from 0 then lmk and I'll make that change. Just wanted to have something that was kind of sensical for people.

not-fl3:
I do not really have a strong opinion here :) 
I am just saying that if the value is not on the X11 - not a big deal, we can invent our own value

narodnik:
OK me neither ;) Will defer to the creator's wisdom here and make the change.

not-fl3:
Thanks for PR!

PR #441, opened 2024-05-31
Fixed cancel_quit on windows not doing anything

Shelim:
At current implementation `cancel_quit` was completely ignored on Windows - the event that could result in preventing quitting the app, was checked after already sending `PostQuitMessage(0)`.

The PR was reasonably tested on Windows in few different configurations. Namely, the `cancel_quit` may acquire `native_display`, so the variable was drop just before entering an event, and restored after. This prevents deadlock in this event.

For easier following the flow, canceling inside `quit_requested_event()` is actually setting `d.quit_requested` to `false` at line 295

### DIFF
--- a/src/native/windows.rs
+++ b/src/native/windows.rs
@@ -284,15 +284,17 @@ unsafe extern "system" fn win32_wndproc(
 
     match umsg {
         WM_CLOSE => {
-            let mut quit_requested = false;
-
             let mut d = crate::native_display().lock().unwrap();
             // only give user a chance to intervene when sapp_quit() wasn't already called
             if !d.quit_ordered {
                 // if window should be closed and event handling is enabled, give user code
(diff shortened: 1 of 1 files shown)


Shelim:
Actually I find a questionable thing inside my implementation: should the `event_handler.quit_requested_event()` be called even if it is the programmer who asked to terminate app? Cancelling shouldn't be possible in such case (or it would break Macroquad `prevent_quit` example, and app would never terminate), but maybe someone want to do some de-intialization on quit, even if it was they who initiate the quit in the first place.

What do you think?

Shelim:
@not-fl3 Any status update on this?

not-fl3:
Ohhh, missed it. I do not have a windows machine to properly test it, tho it looks like it works fine with wine? 

Anyway, lets merge it, will see if there going to be any bug reports!

PR #443, opened 2024-06-03
Implemented blocking_event_loop for osx

birhburh:
Implemented window loop for OSX (as far as I underastand [NSApplication run](https://developer.apple.com/documentation/appkit/nsapplication/1428631-run) works like this)
And also implemented blocking_event_loop for osx (based on this pull request: https://github.com/not-fl3/miniquad/pull/437)


not-fl3:
Thanks for PR!

PR #447, opened 2024-06-12
Macos supress warnings

birhburh:
I'm to blame for most of these warnings, fixed them
And I also runned `cargo fmt`


not-fl3:
Thanks!

PR #448, opened 2024-06-12
Fix "double click" bug on MacOS

eriksvedang:
Here's my attempt at solving the bug described here https://github.com/not-fl3/macroquad/issues/422

The way to reliably (100%) reproduce the bug on MacOS is to launch the program from the command line, and  constantly move the cursor around the screen *while* the program is starting (adding a call to `sleep` at the beginning of the program makes this easier to perform).

This PR uses the method call `[[NSRunningApplication currentApplication] activateWithOptions]` which is documented here: https://developer.apple.com/documentation/appkit/nsrunningapplication/1528725-activatewithoptions

In my testing, the game window is still inactive (gray) after this code has been called, but the double click bug described in the original issue is gone after clicking the window once. That the window actually doesn't get active (which one would assume...) implies that there's more room for improvement here, since the method call actually doesn't make the window active. But the user experience with this fix is at least a lot better than the current behaviour.


birhburh:
Oh, so it was a bug, and not me constantly misclicking.

escobar-west:
I ran into this exact problem today and I am able to reproduce it in my app with the method you suggested (moving the cursor around). I'm going to try build my app with this branch and see if it resolves the problem.  If yes, I'll see what feedback I can provide to help get this (or a similar PR) merged in. Not sure what help I can be besides another tester though.

PR #449, opened 2024-06-12
Implemented blocking_event_loop for Android

birhburh:
And also changed MainActivity.java to compile on older Android


not-fl3:
Thanks for PR!

PR #452, opened 2024-06-17
Implemented blocking_event_loop for ios opengl and metal

birhburh:
- Added MESSAGES_TX for ios and implemented main loop almost similar to android version
- Fixed drawing quarter of screen for metal apple_gfx_api on ios


not-fl3:
Wow! Great PR, thanks!

PR #454, opened 2024-06-17
Bug fix on ios

birhburh:
Follow-up for [this PR](https://github.com/not-fl3/miniquad/pull/452)
It turned out that it isn't redraw after touch without "setNeedDisplay". At least in emulator. Why? I couldn't figure it out. It works ok for Metal. Added something that looks like hack for now.


not-fl3:
Thanks for PR! No problem with a solution that looks like a hack while its well explained in the comment! :+1:

PR #456, opened 2024-06-21
X11 Fix keysym translation using official libxkbcommon

eloraiby:
[x11][xkbcommon] Fix keysym translation using official libxkbcommon instead of bespoke implementation. This should fix the enter to carriage return conversion among other things


not-fl3:
I wonder if it would break miniquad on nixos: https://github.com/not-fl3/miniquad/blob/master/shell.nix

not-fl3:
I wonder if it would break miniquad on nixos: https://github.com/not-fl3/miniquad/blob/master/shell.nix On NixOS each dlopen-able library should be specified in shell.nix.

I dont't see any other real life scenario where libxkbcommon is not available, so I don't think we need to handle a no-libxkb case. But maybe add it to shell.nix to give NixOS/Guix folks a clue that libxkb is a dependency now?

eloraiby:
good observation, seems that libxkbcommon exists in its own domain: https://search.nixos.org/packages?channel=24.05&from=0&size=50&sort=relevance&type=packages&query=libxkbcommon

Added

PR #457, opened 2024-06-22
[GL] Add support for glVertexAttribIPointer (support signed/unsigned byte, short, int attributes)

eloraiby:
Add support for glVertexAttribIPointer (support signed/unsigned byte, short, int attributes).
Add new example that shows support for glVertexAttribIPointer (es3/webgl2/opengl 3+) but needs GLSL 150 to use  OpenGL in MacOS.

### DIFF
--- a/examples/triangle_color4b.rs
+++ b/examples/triangle_color4b.rs
@@ -91,15 +91,19 @@ impl EventHandler for Stage {
 
 fn main() {
     let mut conf = conf::Conf::default();
-    conf.platform.apple_gfx_api = conf::AppleGfxApi::Metal;
-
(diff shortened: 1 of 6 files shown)


not-fl3:
Thanks for PR! 

I think it would be worth it to add a `ContextInfo` flag, maybe `context.info().features.integer_attributes` (not sure about the naming) and to add an explicit panic in the example if that flag is false. 

With miniquad defaulting to WebGl1 on web, that flag being false is a very realistic scenario. Would be nice to have an example checking a platform-specific feature!

eloraiby:
Added as an info function cap.

not-fl3:
Thanks for cleaning up those Info structs, I really like how it looks now! Thanks for PR!

PR #459, opened 2024-06-23
Fix linux_x11 fullscreen toggle

bolphen:
This should allow one to go back from fullscreen mode to windowed mode.


not-fl3:
Thanks!

Issue #455, opened 2024-06-17
MacOS doesn't redraw while window resize is held but not moving


markmurphydev:
Running the Quad example on MacOS on master, the window does not draw while the corner is held, but does draw while dragging it around to resize.

https://github.com/not-fl3/miniquad/assets/2874182/d025d5f3-b855-41f1-9d36-fec2ab27147c

This was introduced in #443

markmurphydev:
I think this is relevant:
https://github.com/rust-windowing/winit/issues/219

markmurphydev:
So the general issue seems to be that when `[ns_app sendEvent:event]` gets a `NSLeftMouseDown` event that would resize the window, it switches into an internal event tracking run loop, which doesn't return control to `macos::run` until resizing is finished.

If you re-add the draw timer from before #443, but set the mode to `NSEventTrackingRunLoopMode`, it will draw while resizing.
I don't really know what the behavior is supposed to be for `blocking_event_loop`, but if you don't want it to draw continuously while resizing, you could create the timer only if it's not blocking.

birhburh:
@markmurphydev, I wrote new implementation (basically copied it from android and ios)
https://github.com/birhburh/miniquad/tree/blocking_event_loop_osx
It is still raw but should work already

markmurphydev:
I noticed that the window also freezes while any drop-down on the osx top menu bar is open, but this is also fixed by #462

PR #460, opened 2024-06-24
Reimplemented blocking_event_loop for OSX

birhburh:
While solving [this problem](https://github.com/not-fl3/miniquad/issues/455), decided to take a less hacky approach

- Reverted own NSApplication run implementation
- Used same approach as for android and ios
- Catched the bug with examples/post_processing.rs turning black, checked that it existed even before own NSApplication run implementation


birhburh:
Ok
Find out that Metal backend isn't working yet

PR #463, opened 2024-06-25
graphics: Fix backwards compatibility with integer attributes

not-fl3:
Follow up on https://github.com/not-fl3/miniquad/pull/461

I can't really think of any better solution, just added a doc comment over `gl_pass_as_float`. 

cc @eloraiby @birhburh

### DIFF
--- a/src/graphics/gl.rs
+++ b/src/graphics/gl.rs
@@ -738,10 +738,6 @@ impl GlContext {
         unsafe { glColorMask(r as _, g as _, b as _, a as _) }
         self.cache.color_write = color_write;
     }
-
-    fn has_integer_attributes(&self) -> bool {
-        self.info.has_integer_attributes()
-    }
 }
 
 fn gl_info(features: Features) -> ContextInfo {
@@ -1070,6 +1066,7 @@ impl RenderingBackend for GlContext {
             name,
             format,
             buffer_index,
+            gl_pass_as_float,
         } in attributes
         {
             let buffer_data = &mut buffer_cache
@@ -1105,6 +1102,7 @@ impl RenderingBackend for GlContext {
                         stride: buffer_data.stride,
                         buffer_index: *buffer_index,
                         divisor,
+                        gl_pass_as_float: *gl_pass_as_float,
                     };
 
                     assert!(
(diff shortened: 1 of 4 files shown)


src/graphics/gl.rs
```diff
@@ -1332,7 +1334,9 @@ impl RenderingBackend for GlContext {
                     unsafe {
                         match attribute.type_ {
                             GL_INT | GL_UNSIGNED_INT | GL_SHORT | GL_UNSIGNED_SHORT
-                            | GL_UNSIGNED_BYTE | GL_BYTE => {
+                            | GL_UNSIGNED_BYTE | GL_BYTE
+                                if !attribute.gl_pass_as_float =>
```
eloraiby, line 1338: QQ: why not something like:
```
                        match attribute.type_ {
                            GL_INT | GL_UNSIGNED_INT | GL_SHORT | GL_UNSIGNED_SHORT
                            | GL_UNSIGNED_BYTE | GL_BYTE
                                if self.has_integer_attributes() =>
                            {
                                glVertexAttribIPointer(
                                    attr_index as GLuint,
                                    attribute.size,
                                    attribute.type_,
                                    attribute.stride,
                                    attribute.offset as *mut _,
                                )
                            }
                            _ => glVertexAttribPointer(
                                attr_index as GLuint,
                                attribute.size,
                                attribute.type_,
                                GL_FALSE as u8,
                                attribute.stride,
                                attribute.offset as *mut _,
                            ),
                        }
```

not-fl3: thx, I agree, a lot better!

not-fl3: oh, wait, I misread  it. 

`has_integer_attributes` doesn't neccessarily mean we want uvec4 in the shader. 

`has_integer_attributes` means the backend support glVertexAttribIPointer, `gl_pass_as_float` means that the user wants glVertexAttribIPointer to be used.

eloraiby: I don't think you d need `gl_pass_as_float`, because the moment it supports integer and the user is passing a byte, it'll convert it. glsl 150 requires opengl 3.2 and above, so that shouldn't affect it. But better test this one.

not-fl3: It does not, thats how that whole thing started :) 
glsl100 shaders do not receive glVertexAttribIPointer Byte4 as vec4, even on OpenGL4+

not-fl3:
triangle_color4b example works, macroquad works, all old miniquad using code should not be affected, so lets merge it and will see how it goes! 

Feel free to send follow up PRs if it was not, in fact, a good idea :)

PR #464, opened 2024-06-29
Add a quit shortcut to the osx menu bar 

markmurphydev:
Adds Command+Q functionality by adding a "Quit AppName" entry to the top menu bar. This is the normal way to implement the command-q shortcut on osx, and lets you rebind it in system keyboard settings.

Fixes not-fl3/macroquad#649

Note that on osx on master, clicking on a menu bar item currently freezes the window. This is due to #455.

### DIFF
--- a/src/native/macos.rs
+++ b/src/native/macos.rs
@@ -907,31 +907,31 @@ unsafe fn initialize_menu_bar(ns_app: ObjcId) {
     // Adapted from Winit `menu::initialize`
 
     // System menu bar
-    let menu_bar: ObjcId = msg_send![class!(NSMenu), new];
+    let menu_bar = msg_send_![class!(NSMenu), new];
     // Entry for the app menu dropdown in the menu bar
-    let app_menu_item: ObjcId = msg_send![class!(NSMenuItem), new];
-    let app_menu: ObjcId = msg_send![class!(NSMenu), new];
+    let app_menu_item = msg_send_![class!(NSMenuItem), new];
+    let app_menu = msg_send_![class!(NSMenu), new];
 
     // Hook up the menu components to the application
-    let () = msg_send![app_menu_item, setSubmenu: app_menu];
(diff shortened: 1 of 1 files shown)


src/native/macos.rs
```diff
@@ -901,6 +901,40 @@ unsafe fn set_icon(ns_app: ObjcId, icon: &Icon) {
     CGImageRelease(image);
 }
 
+/// Initialize the system menu bar for this application
+/// - ns_app: This NSApplication
+#[no_mangle]
```
not-fl3, line 906: does it really need to be `no_mangle`?

src/native/macos.rs
```diff
@@ -901,6 +901,40 @@ unsafe fn set_icon(ns_app: ObjcId, icon: &Icon) {
...
+    // Adapted from Winit `menu::initialize`
+
+    // System menu bar
+    let menu_bar: ObjcId = msg_send![class!(NSMenu), new];
+    // Entry for the app menu dropdown in the menu bar
+    let app_menu_item: ObjcId = msg_send![class!(NSMenuItem), new];
+    let app_menu: ObjcId = msg_send![class!(NSMenu), new];
+
+    // Hook up the menu components to the application
+    let () = msg_send![app_menu_item, setSubmenu: app_menu];
```
not-fl3, line 917: up to you, but there is a special macro with a dash, `msg_send_![app_menu_item, setSubmenu: app_menu];`

src/native/macos.rs
markmurphydev: Definitely not. I was messing with the debugger and forgot to remove that.

src/native/macos.rs
markmurphydev: Oh that's nicer looking. Switched to using that.

not-fl3:
Thanks for PR!

PR #466, opened 2024-07-04
native/linux_wayland: add missing insert_char event

narodnik:
We do this by using keysyms for WaylandEvent::KeyboardKey directly.


not-fl3:
Oh, somehow I missed it, sorry for delay! 

Thanks for PR!

PR #471, opened 2024-08-07
Add support for mouse buttons 4 & 5

guy-bartkus:
Trying to add support for mouse buttons 4 & 5 to [macroquad](https://github.com/not-fl3/macroquad).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,9 @@ categories = ["rendering::graphics-api"]
 # disabled by default
 log-impl = []
 
+[dependencies]
+num_enum = "0.7.3"
+
 [target.'cfg(target_os = "linux")'.dependencies]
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -20,6 +20,9 @@ categories = ["rendering::graphics-api"]
 # disabled by default
 log-impl = []
 
+[dependencies]
+num_enum = "0.7.3"
```
not-fl3, line 24: `num_enum` adds full rust parser and a bunch of other dependencies into the build try. I would rather manually implement TryFrom for one enum we need it for.

guy-bartkus: Okay, I used a conversion function instead to convert u8 to MouseButton type. Is this better?

PR #472, opened 2024-08-07
add default filter mode to conf

0x177:
this is a continuation of [this macroquad pull request](https://github.com/not-fl3/macroquad/pull/772)

### DIFF
--- a/src/conf.rs
+++ b/src/conf.rs
@@ -182,6 +182,11 @@ pub struct Conf {
     /// Platform specific settings. Hints to OS for context creation, driver-specific
     /// settings etc.
     pub platform: Platform,
+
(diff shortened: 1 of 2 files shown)


not-fl3:
I do not think this belongs to miniquad, macroquad's Conf would do just fine

0x177:
in the [provided api example](https://github.com/not-fl3/macroquad/pull/772#issuecomment-2271639233), default_filter_mode is a field in a miniquad Conf

not-fl3:
I do not think miniquad have anything to do with macroquad's default texture filters. 

I would rather have it in this struct, https://docs.rs/macroquad/latest/macroquad/conf/struct.Conf.html

Issue #470, opened 2024-08-04
MacOS CPU usage increased by 8x in v0.4.4


brettchalupa:
I noticed running a simple Macroquad example on MacOS 14.5 with an Apple M2 Pro chip was using a high amount of CPU. I checked out some previous versions of Macroquad and saw the performance was much better in older versions. Using `git bisect`, I landed on https://github.com/not-fl3/macroquad/commit/93b8af24fd99d8a8f2cd4e2446260a43df36e1d6 being the introduction of the high CPU usage,  which contains a patch level upgrade for miniquad from 0.4.2 to 0.4.3 (which was yanked) and macroquad_macro from 0.1.7 to 0.1.8. 

So I ran a `git bisect` on Miniquad, using `cargo run --release --example triangle` as my test for CPU usage. Some examples:

- v0.4.0, 36dd9d911ebe818a3adf2da82e7f3f0530291dae - CPU @ 11% for `cargo run --release --example triangle`
- v0.4.5, a962bfd9781e431335989775fd42335b79317397 - CPI @ 80% for `cargo run --release --example triangle`

Through the bisect, I landed on and confirmed that this commit https://github.com/not-fl3/miniquad/commit/833799ec32883cdd7465a5f8ddc80e2203dbcbc4 increases CPU load on MacOS by about 8x. First noticeable with the v0.4.4 release of Miniquad.

## Repro steps

1. Clone the repo
2. Checkout the commit just before: `git checkout 833799ec32883cdd7465a5f8ddc80e2203dbcbc4^1`
3. Run `cargo run --release --example triangle` to see CPU performance baseline
4. Checkout the high CPU usage commit: `git checkout 833799ec32883cdd7465a5f8ddc80e2203dbcbc4`
5. Run `cargo run --release --example triangle` to see CPU performance degrade

## Additional info

Happy to help test or debug, especially if access to MacOS is limited. @birhburh pinging you here too since you authored the commit. Let me know if I can be supportive in any way. Thanks!

birhburh:
@brettchalupa, Thanks!
Yes, checked and this is also true for me on Intel
I'll look into this

VanjaRo:
It seems that changin the config field `blocking_event_loop`to `true` for the macos fixes the problem. As I mention in the PR, maybe `run` method of the NSApplication class was utilizing the same mecanism under the hood.

VanjaRo:
[...]

PR #475, opened 2024-08-11
native/macos: Re-re-implemented event loop

birhburh:
- Reverted macos.rs to this implementation: https://github.com/not-fl3/miniquad/pull/443 (it's probably better compare changes with this PR than with latest commit to understand what was added additionally)
- Now using own NSView with NSOpenGLContext instead of NSOpenGLView
- For metal backend using redraw instead setNeedsDisplay, because somehow it reduces cpu usage (Cannot find info about enabling vsync like this in MTKView docs)
- Fixed freezing on resize by drawing in draw_rect that called during "live resize"™. I don't like this approach, but it blocks main event loop while resizing, so it will not be some kind of concurency of opengl stuff i think
- Reducing CPU usage when window is occluded
- Added comments to hacky places, there are lots of them imo

Fixes:
- https://github.com/not-fl3/miniquad/issues/455
- https://github.com/not-fl3/miniquad/issues/470

### DIFF
--- a/src/native/apple/frameworks.rs
+++ b/src/native/apple/frameworks.rs
@@ -75,6 +75,7 @@ extern "C" {
 extern "C" {
     pub static NSRunLoopCommonModes: ObjcId;
     pub static NSDefaultRunLoopMode: ObjcId;
+    pub static NSEventTrackingRunLoopMode: ObjcId;
     pub static NSProcessInfo: ObjcId;
     pub fn NSStringFromClass(class: ObjcId) -> ObjcId;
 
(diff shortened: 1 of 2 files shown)


not-fl3:
Can't say that I understand a problem enough to have an opinion on the high-level approach, but reading through the code nothing strikes me wrong. I did not found excess of hacks, and with proper comments they did not really looked like too much of hacks anyway. Well, I guess it counts as a review. 

In other words, great work, I trust your judgement here, @birhburh, lets merge it and will see whats up with bug reports!

brettchalupa:
I tested out this latest code on `master` on my Apple Silicon chip on macOS and it segfaults when running `cargo run --release --example triangle`. Possibly the same as this segfault on @birhburh's fork's prototype branch https://github.com/not-fl3/miniquad/issues/470#issuecomment-2284167669 but not 100% sure.

Happy to provide some more details or test things out. Let me know if a new issue is desired. Thanks!

PR #476, opened 2024-08-12
native/macos: Fixes setNeedsDisplay call on arm macs

birhburh:
Possible fix for: https://github.com/not-fl3/miniquad/issues/470#issuecomment-2284167669
> If I had a nickel for every time I needed to use setNeedsDisplay hack, I'd have two nickels. Which isn't a lot, but it's weird that it happened twice

@brettchalupa, can you please recheck now
And with metal also, please, just to be sure (`cargo run --example quad metal`)

### DIFF
--- a/src/native/macos.rs
+++ b/src/native/macos.rs
@@ -612,7 +612,6 @@ unsafe fn view_base_decl(decl: &mut ClassDecl) {
         payload.modifiers = new_modifiers;
     }
 
-
     decl.add_method(
         sel!(canBecomeKey),
         yes as extern "C" fn(&Object, Sel) -> BOOL,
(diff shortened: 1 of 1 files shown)


src/native/macos.rs
```diff
@@ -712,6 +711,18 @@ pub fn define_opengl_view_class() -> *const Class {
         }
     }
 
+    // apparently, its impossible to use performSelectorOnMainThread
+    // with primitive type argument, so the only way to pass
+    // YES to setNeedsDisplay - send a no argument message
+    // https://stackoverflow.com/questions/6120614/passing-primitives-through-performselectoronmainthread
+    // It seems that the same thing applies to [NSTimer timerWithTimeInterval:...]
+    extern "C" fn set_needs_display_hack(this: &Object, _: Sel) {
```
not-fl3, line 719: shame on you, apple, I wish we did not need to do this :(

brettchalupa:
@birhburh confirmed that `cargo run --release --example triangle` runs on macOS on Silicon and does not seg fault. Also CPU usage is much lower, at about 8% with that example.

I confirmed `cargo run --example quad metal` runs as well and uses lower CPU usage.

Nice! 🙌

PR #479, opened 2024-09-04
Resizing the canvas using js/css now triggers the resize event

yui-915:
fixes [macroquad#787](https://github.com/not-fl3/macroquad/issues/787) (I assume)

### DIFF
--- a/js/gl.js
+++ b/js/gl.js
@@ -420,13 +420,14 @@ function _webglGet(name_, p, type) {
 var Module;
 var wasm_exports;
 
-function resize(canvas, on_resize) {
+function resize(canvas, on_resize, force = false) {
     var dpr = dpi_scale();
     var displayWidth = canvas.clientWidth * dpr;
     var displayHeight = canvas.clientHeight * dpr;
 
     if (canvas.width != displayWidth ||
-        canvas.height != displayHeight) {
+        canvas.height != displayHeight ||
+        force) {
         canvas.width = displayWidth;
         canvas.height = displayHeight;
         if (on_resize != undefined)
@@ -1263,6 +1264,9 @@ var importObject = {
                     wasm_exports.touch(SAPP_EVENTTYPE_TOUCHES_MOVED, touch.identifier, relative_position.x, relative_position.y);
                 }
             });
+            new ResizeObserver(function (_entries) {
+                resize(canvas, wasm_exports.resize, true);
+            }).observe(canvas);
 
             window.onresize = function () {
                 resize(canvas, wasm_exports.resize);


InnocentusLime:
As the original issue creator, I have done some local tests and now the application **DOES** seem to be responding to me changing the canvas size in the browser! :D

I have tested this on both Chrome and Firefox. But I have no way to test it on Safari

js/gl.js
```diff
@@ -420,13 +420,14 @@ function _webglGet(name_, p, type) {
 var Module;
 var wasm_exports;
 
-function resize(canvas, on_resize) {
+function resize(canvas, on_resize, force = false) {
```
not-fl3, line 423: I can't see this `force` flag being used, am I mising something?

js/gl.js
```diff
@@ -420,13 +420,14 @@ function _webglGet(name_, p, type) {
...
     var dpr = dpi_scale();
     var displayWidth = canvas.clientWidth * dpr;
     var displayHeight = canvas.clientHeight * dpr;
 
     if (canvas.width != displayWidth ||
-        canvas.height != displayHeight) {
+        canvas.height != displayHeight ||
+        force) {
```
not-fl3, line 430: I am concerned about this `force` flag. Why do we need to send a resize message if size is still the same?

js/gl.js
```diff
@@ -1263,6 +1264,9 @@ var importObject = {
                     wasm_exports.touch(SAPP_EVENTTYPE_TOUCHES_MOVED, touch.identifier, relative_position.x, relative_position.y);
                 }
             });
+            new ResizeObserver(function (_entries) {
```
not-fl3, line 1267: if we use both `onresize` and `ResizeObserver`, wouldn't it generate resize message twice?

js/gl.js
lenscas: haven't looked at the code so I might be wrong

but perhaps it is a way to update the size if the CSS or similar changes?

not-fl3: Maybe I am missing something, but I can't see how any sort of resize can be meaningful if  canvas.width/height is still the same?

js/gl.js
yui-915: yes .... especially that `ResizeObserver` is triggered after `window.onresize`, and it has that `force` flag

simply removing `window.onresize` would solve it
but it'll cause the canvas to blink black when resizing the window, probably due to the order things are done in the browser's event loop and the fact that updating the canvas size clears it 

the best solution I found was changing `ResizeObserver` to the following:
```js
new ResizeObserver(() => {
    let dpr = dpi_scale();
    let displayWidth = Math.floor(canvas.clientWidth * dpr);
    let displayHeight = Math.floor(canvas.clientHeight * dpr);
    if (wasm_exports.resize)
        wasm_exports.resize(displayWidth, displayHeight);
}).observe(canvas);
```
and removing the `on_resize` callback from the `resize` function 
(also the `force` flag as it's no longer needed)

js/gl.js
yui-915: as far as my tests went

when forcing the canvas to be "fullscreen"
resizing the browser window will change `canvas.clientWidth`, but not `canvas.width`

but when removing `width: 100%` and `height: 100%` from the canvas css
manually updating the canvas width using js (for example by doing `canvas.width += 10;`)
will change both `canvas.width` and `canvas.clientWidth` 
so the condition evaluates to false, and on_resize method doesn't get called

same applies for height

js/gl.js
yui-915: something like yui-915/miniquad@53b457e4

PR #480, opened 2024-09-08
Fix `target_ios` typo

cyrgani:
Fixes a simple typo: `#[cfg(target_ios = "ios")]` becomes `#[cfg(target_os = "ios")]`.


not-fl3:
Thanks!

PR #482, opened 2024-09-14
Add get/set_window_position for Linux X11

InZan17:
Adds functionality to `get_window_position` and `set_window_position` functions for Linux X11, which was left out on #434.


not-fl3:
:+1: 
Thanks for PR!

PR #491, opened 2024-10-02
replace usage of the `std::os::raw` module with `core::ffi`

cyrgani:
`std::os::raw` is semi-deprecated since six years and recommends using `core::ffi` instead.
This PR is just a replacement of all usages of this module.


not-fl3:
Thanks for PR!

PR #496, opened 2024-11-23
make module native public

narodnik:
I'm following this guide https://macroquad.rs/articles/java/

which requires `miniquad::native::android::{self, ndk_sys, ndk_utils}` to be public in order to extend the Android java files.


not-fl3:
Makes sense, thanks for PR!

PR #499, opened 2024-11-30
When the IME is shown, it covers the systemBars at the bottom, so just use IME height for the bottom padding

narodnik:


not-fl3:
Thanks for PR!

PR #503, opened 2024-12-12
add warning messages when wayland fails to load due to shared libraries

Adjective-Object:
While trying to debug why macroquad was failing to run in a container, I ran into a variation of [this issue](https://github.com/not-fl3/macroquad/issues/658), which turned out to just be that I was missing `libwayland-egl` in the container.

This would have been more straightforward to debug if miniquad printed the reason it failed to initialize when it fails, which it does in a few other scenarios.

This change adds eprintln! warnings when a required library is missing

### DIFF
--- a/src/native/linux_wayland.rs
+++ b/src/native/linux_wayland.rs
@@ -533,14 +533,22 @@ impl crate::native::Clipboard for WaylandClipboard {
     fn set(&mut self, _data: &str) {}
 }
 
+
+pub fn eprint_iferr<T>(lib_name: &'static str, opt_lib: Option<T>) -> Option<T> {
+    if opt_lib.is_none() {
+        eprintln!("failed to load library {}", lib_name);
+    }
+    return opt_lib
+}
+
 pub fn run<F>(conf: &crate::conf::Conf, f: &mut Option<F>) -> Option<()>
 where
     F: 'static + FnOnce() -> Box<dyn EventHandler>,
 {
     unsafe {
-        let client = LibWaylandClient::try_load()?;
-        let egl = LibWaylandEgl::try_load()?;
-        let xkb = LibXkbCommon::try_load()?;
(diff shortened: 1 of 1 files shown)


Adjective-Object:
For reference I wasn't able to fully test this -- it looks like there's some unrelated failing tests in `cargo test`?
image

src/native/linux_wayland.rs
```diff
@@ -533,14 +533,22 @@ impl crate::native::Clipboard for WaylandClipboard {
...
+
 pub fn run<F>(conf: &crate::conf::Conf, f: &mut Option<F>) -> Option<()>
 where
     F: 'static + FnOnce() -> Box<dyn EventHandler>,
 {
     unsafe {
-        let client = LibWaylandClient::try_load()?;
-        let egl = LibWaylandEgl::try_load()?;
-        let xkb = LibXkbCommon::try_load()?;
+        let client = eprint_iferr("libwayland-client", LibWaylandClient::try_load())?;
```
not-fl3, line 549: Maybe try_load could return a better message instead of handling this here?

Adjective-Object: for sure, it would be more natural for the various Lib___::try_load() functions to return Result<>s rather than Option<>s

Adjective-Object:
Verified that the latest round of changes i pushed still works by running within a docker container with missing dependencies:
```
warning: `game` (bin "cool_game") generated 2 warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.62s
     Running `target/debug/cool_game`
failed loading libwayland-egl: failed opening library libwayland-egl.so.1
Failed to initialize through wayland! Trying X11 instead
Authorization required, but no authorization protocol specified

thread 'main' panicked at /workspaces/miniquad/src/native/linux_x11.rs:585:13:
XOpenDisplay() failed!
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

PR #512, opened 2025-01-10
macos: add window_resizable impl

kotx:
Resolves not-fl3/macroquad#210 on macOS only.
This makes `window_masks` mutable but it can be changed to be within a block instead.

Let me know if I'm missing something obvious like a set_resizable function, it's my first time looking at this codebase.


not-fl3:
Thanks for PR!

PR #515, opened 2025-01-16
linux_wayland: fix bug where shift+i, release shift, release i sends repeated I events

narodnik:
because the key -> keysym translation applies modifiers which confuses the repeated_keys hashset. Instead we should use the raw key itself, and apply the modifiers when forwarding them to the API.


not-fl3:
Thanks for PR!

Issue #423, opened 2024-04-10
EventHandler::files_dropped_event not usable due to Window not allowing dropping files


Nrosa01:
I trying out miniquad to see if I could port the EventHandler::files_dropped_event to macroquad but it seems it won't work even in miniquad. By reading the codet the event seems to be correctly managed, so it should be possible to use it:

src/lib.rs

```rust
189    pub fn dropped_file_count() -> usize {
190        let d = native_display().lock().unwrap();
191        d.dropped_files.bytes.len()
192    }
193    pub fn dropped_file_bytes(index: usize) -> Option<Vec<u8>> {
194        let d = native_display().lock().unwrap();
195        d.dropped_files.bytes.get(index).cloned()
196    }
197    pub fn dropped_file_path(index: usize) -> Option<std::path::PathBuf> {
198        let d = native_display().lock().unwrap();
199        d.dropped_files.paths.get(index).cloned()
200    }
```

Now, the important, when you try to drop something to a miniquad windows:

![image](https://github.com/not-fl3/miniquad/assets/62796415/3392d503-767d-4df4-a1eb-ef8266324ffe)

You can see the windows block cursor icon that appears when I drag a file from the explorer onto the windows. Even if miniquad functions are completed an able to work, the Windows itself won't allow it.

I wanted to contribute to help fix this as it's a feature important for my project. I've been reading miniquad codebase for a while but I don't get why this happens.

I'll continue reading the code and trying to make this work cross-platform. There are some similar issues so there might be more people affected by this, but those issues are old and doesn't explain if the function itself wasn't working or if it was the same problem exposed here.

not-fl3:
related PR that I believe was not fully migrated to miniquad-0.4 https://github.com/not-fl3/miniquad/pull/275

Nrosa01:
Oh I didn't see that PR.Thank you! I'll try to migrate it to miniquad 0.4

Nrosa01:
I found out this works on wasm, it seems to fail on Windows. Sadly I don't have a Linux machine to test, I'll install vmware player once I get this working on windows and check if it works on Linux

Issue #274, opened 2022-04-14
File drag & drop support


urholaukkarinen:
I was building an app using egui-macroquad and noticed that it does not support file drag & drop. Seems like it is not implemented in sapp. Have there been plans to implement this? I would be able to work on the Windows, Mac, Linux and Wasm implementations.

brainwo:
Any update on this?

PR #518, opened 2025-01-24
Implement winapi dropping files

zyllian:
Implements drag and drop file support for Miniquad Windows using winapi.

Note that this implementation crashes if a folder is dropped onto the window as I didn't want to refactor any Miniquad APIs as part of this PR.

Related issues: #274, #423


not-fl3:
Thanks for PR!

Issue #517, opened 2025-01-24
Fix broken keyboard input API


narodnik:
Proposed API:

```
fn key_event(KeyEvent, KeyPhase);
```

KeyEvent then has this API: https://doc.qt.io/qt-6/qkeyevent.html

KeyPhase has Press, Release

Check this log:

```
07:22:12 [DEBUG] (1) gfx::event: char_event(", KeyMods { shift: true, ctrl: false, alt: false, logo: false }, false)
07:22:12 [DEBUG] (1) gfx::event: char_event(", KeyMods { shift: true, ctrl: false, alt: false, logo: false }, true)
07:22:12 [DEBUG] (1) gfx::event: char_event(", KeyMods { shift: true, ctrl: false, alt: false, logo: false }, true)
07:22:13 [DEBUG] (1) gfx::event: char_event(', KeyMods { shift: false, ctrl: false, alt: false, logo: false }, true)
07:22:13 [DEBUG] (1) gfx::event: char_event(', KeyMods { shift: false, ctrl: false, alt: false, logo: false }, true)
07:22:13 [DEBUG] (1) gfx::event: char_event(', KeyMods { shift: false, ctrl: false, alt: false, logo: false }, true)
```

This is the same key being held, but the API does not give any indication of that.

The proposed API would fix this, since KeyEvent gives access to the native virtual keycodes for that platform. They should not be relied upon to be consistent across platforms but to simply map to unique keys on that device. This would allow us to clearly see key down and release events despite the char/key changing.

Also you can call .chr() to get the actual char, or try to get the KeyCode. This would make existing code simpler too, since right now I have code like:

```rust
/// Filter these char events from being handled since we handle them
/// using the key_up/key_down events.
/// Avoids duplicate processing of keyboard input events.
pub static DISALLOWED_CHARS: &'static [char] = &['\r', '\u{8}', '\u{7f}', '\t', '\n'];

/// These keycodes are handled via normal key_up/key_down events.
/// Anything in this list must be disallowed char events.
pub static ALLOWED_KEYCODES: &'static [KeyCode] = &[
    KeyCode::Left,
    KeyCode::Right,
    KeyCode::Up,
    KeyCode::Down,
    KeyCode::Enter,
    KeyCode::Kp0,
    KeyCode::Kp1,
    KeyCode::Kp2,
    KeyCode::Kp3,
    KeyCode::Kp4,
    KeyCode::Kp5,
    KeyCode::Kp6,
[...]

PR #519, opened 2025-01-26
clear the modifier keys on exiting the focused window

narodnik:
we need to clear the mods otherwise they can be released outside the window's focus and stay on when the window is refocused.

btw when you get a chance, please approve issue https://github.com/not-fl3/miniquad/issues/517, ty!

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -181,6 +181,8 @@ pub trait EventHandler {
 
     fn key_down_event(&mut self, _keycode: KeyCode, _keymods: KeyMods, _repeat: bool) {}
 
+    /// Note: you are not always guaranteed to receive a key_up event. For example on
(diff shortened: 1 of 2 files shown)


not-fl3:
Thanks!

profan:
has this only fixed this under the wayland backend and only on linux, or was this a wayland-only problem?

i'm pretty sure there's this exact issue but for macroquad, which is likely just originating in miniquad: https://github.com/not-fl3/macroquad/issues/547

happens on windows as well, clearing the mods when exiting the focused window also doesn't seem quite correct, other libraries like SDL2 or the like properly handle releasing keys when you've moused off the window, so this feels and looks like a little bit of a bandaid fix that will not actually fix the problem?

PR #520, opened 2025-01-26
Android: Adding FULLSCREEN flag when the window is fullscreen.

MohammedKHC0:
This flag is needed to properly hide display cutout on Android devices,
Without it a black space appears on the status bar.
Refer to
https://stackoverflow.com/questions/49190381/fullscreen-app-with-displaycutout


MohammedKHC0:
Hmm. You are right about that, it's deprecated. What i would typically do to hide everything is to use WindowInsetsController*Compat*. But because MiniquadActivity is not a subclass of AppCompatActivity, I think the best solution is to check if api >= 30, and then hide display cutout and status bar using WindowInsetsController..
I changed the code based on what WindowInsetsContoller do. Please review it.

PR #525, opened 2025-02-03
replace `OnceLock` with `AtomicPtr` in `NATIVE_DISPLAY`

joseluis:
I wanted to be able to start and stop a miniquad context and after trying several solutions to make `NATIVE_DISPLAY` droppable (e.g. using `RwLock`), I believe this one with `AtomicPtr` plus an static `AtomicBool` has a very good balance between performance and security.

A simple test:
```rust
use miniquad::*;

struct Stage {
    ctx: GlContext,
}
impl EventHandler for Stage {
    fn update(&mut self) {}
    fn draw(&mut self) {}
}
fn main() {
    println!("first context\nclose the window to create a second context.");
    miniquad::start( conf::Conf::default(), || { Box::new(Stage { ctx: GlContext::new(), }) });
    println!("second context");
    miniquad::start( conf::Conf::default(), || { Box::new(Stage { ctx: GlContext::new(), }) });
}
```

I also got the idea of the `native_display` function wrappers from #505 to improve ergonomics.

### DIFF
--- a/src/graphics/gl.rs
+++ b/src/graphics/gl.rs
@@ -520,6 +520,11 @@ pub struct GlContext {
     pub(crate) cache: GlCache,
     pub(crate) info: ContextInfo,
 }
+impl Drop for GlContext {
+    fn drop(&mut self) {
+        crate::drop_display();
+    }
+}
 
 impl GlContext {
     pub fn new() -> GlContext {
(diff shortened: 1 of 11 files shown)


narodnik:
You have a Mutex, so it makes no sense to wrap the Mutex in an AtomicPtr. Why not just do this?

```rust
/// This for now is Android specific since the process can continue running but the display
/// is restarted. We support reinitializing the display.
fn set_or_replace_display(display: native::NativeDisplayData) {
    if let Some(m) = NATIVE_DISPLAY.get() {
        // Replace existing display
        *m.lock().unwrap() = display;
    } else {
        // First time initialization
        set_display(display);
    }
}
```

PR #530, opened 2025-02-12
osx: Fix calling [window center] before setting gl_context

birhburh:
For
```sh
cargo +nightly run --example=quad                  
```
I recieved
```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/examples/quad`

thread 'main' panicked at src/native/macos.rs:304:13:
null pointer dereference occurred
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread caused non-unwinding panic. aborting.
zsh: abort      cargo +nightly run --example=quad
```

Also decided to move some functions below gl_context initialization just to be sure that everything is inited before creation of view and window
This probably could fix some other osx bugs with mouse input

---
Why I needed nightly compiler for miniquad is because I tried to use glam with simd feature, but it is probably not important

### DIFF
--- a/src/native/macos.rs
+++ b/src/native/macos.rs
@@ -1079,8 +1079,6 @@ where
     let title = str_to_nsstring(&conf.window_title);
     //let () = msg_send![window, setReleasedWhenClosed: NO];
     let () = msg_send![window, setTitle: title];
-    let () = msg_send![window, center];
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks for PR!

PR #531, opened 2025-02-15
native/linux_x11: INCR and Drag-n-Drop

bolphen:
- Refactor clipboard code and implement INCR protocol for receiving
- Implement Drag-n-Drop

Works on my machine™ :)


not-fl3:
Great work, thanks for PR!

Issue #501, opened 2024-12-09
Linux/X11 can't display window icon properly


Garfield-hr:
Hi,

Thank you for developing this remarkable render engine.

I'm currently developing a simulation software based on miniquad on Linux. I noticed that window icon support for Linux/X11 was added in this PR, and I'm trying to implement an icon for my application.

However, despite the merged changes, I'm still unable to see any window icon on my Linux/X11 system. I've tested this with the triangle example, and the window still does not show icon (please see the attached screenshot).
![image](https://github.com/user-attachments/assets/9923dfc6-bfb1-46e4-a2cc-4667f4857915)

Based on my debugging:

- I've verified that the update_window_icon function is being called with the proper icon data (confirmed via debug prints)
- On windows icon can be displayed properly
- On my Linux/X11 system, no window icon appears
Could you help confirm if there are any additional X11-specific steps needed to properly implement the window icon functionality?

Below are the specifications of my system:
- OS: Ubuntu 20.04.6 LTS
- XDisplay:
  - version number: 11.0
  - X.org version: 1.20.13

Thank you for your time.

PR #532, opened 2025-02-19
native/linux_x11: Allow setting WM_CLASS

bolphen:
Following #501

Since `Platform` is `Copy`, I used `&'static str` instead of `String` for the name. Shouldn't be a problem cuz you won't want to change it.

Also fixed a typo in icon :)


not-fl3:
Thanks for PR! 
Interesting that this is the requirement for gnome, but if it works - it works, I guess!

PR #533, opened 2025-02-26
native/linux: Use macro to reduce boilerplate

bolphen:
This avoids writing the same function name 5 times in 3 different places.

~~I didn't convert the egl file yet.~~ Did the egl file as well. The gl one seems to have its own macro.

The macro itself is inspired by the dlib crate, but I made it less verbose with the `...` syntax. I'm also open to other suggestions.

Once this got merged I'll start submitting the PRs for the Wayland backend :)

### DIFF
--- a/src/native/android.rs
+++ b/src/native/android.rs
@@ -142,13 +142,13 @@ struct MainThreadState {
 
 impl MainThreadState {
     unsafe fn destroy_surface(&mut self) {
-        (self.libegl.eglMakeCurrent.unwrap())(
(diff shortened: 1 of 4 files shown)


not-fl3:
Great work! :+1: 

I am little bit on the fence about `...`, but can't really build a strong opinion on this. Intuitively explicit `static`/`function` etc lists would be easier to read, but not by much and we only use the macro a handful of times. I do not have very strong opinion on this tbh, very excited to see how it will turn out with wayland!

PR #534, opened 2025-02-26
actually use dpi_scale() on wasm backend

kampffrosch94:
I was messing around with scaling in wasm builds, trying to get pixel perfect results.
I was pretty confused why `screen_dpi_scale()` in macroquad always reported 1.0 even when that was clearly wrong.
grafik

It turns out that the DPI value was never written to the NativeDisplayData.
This PR fixes that.


not-fl3:
Good catch, thanks for PR!

PR #535, opened 2025-02-28
Wayland backend

bolphen:
We now have
- Cursor shape / hide and grab (the cursor shape protocol is not supported under gnome for now but apparently it will be there in the next version)
- Key repeat
- Clipboard and drag-n-drop
- Blocking event loop
- High-dpi support
- Client-side window decorations under gnome using `libdecor`, which is what xwayland uses (I removed the old fallback decorations: it looks bad but more importantly it also messes up the cursor positioning, and I don't want to manually fix the offset)

Probably some tests are needed before merging, especially under different compositors and on different machines (I tested on `mutter`, `hyprland`, and `sway`, and I've had the same code working under one but not another...)

Also I did modify the code structure quite a bit. Probably gonna be bad for others who worked/are working on it (@narodnik maybe? btw I rebased upon your clipboard implementation; it got me started so your work is not in vain :)

There's still some code that's a bit messy; also it'd probably be nice for the X11 and Wayland implementations to share more common patterns, which I didn't pay enough attention. But at least it works now :)

### DIFF
--- a/src/conf.rs
+++ b/src/conf.rs
@@ -101,6 +101,22 @@ pub enum WebGLVersion {
     WebGL2,
 }
 
+/// On Wayland, specify how to draw client-side decoration (CSD) if server-side decoration (SSD) is
+/// not supported (e.g., on GNOME).
+///
+/// Defaults to ServerWithLibDecorFallback
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Hash)]
+pub enum WaylandDecorations {
+    /// If SSD is not supported, will try to load `libdecor` to draw CSD. This is the default
+    /// choice.
+    #[default]
+    ServerWithLibDecorFallback,
+    /// If SSD is not supported, draw a light gray border.
+    ServerWithMiniquadFallback,
+    /// If SSD is not supported, no CSD will be drawn.
+    ServerOnly,
+}
+
 /// Platform-specific settings.
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 pub struct Platform {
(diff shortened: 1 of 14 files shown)


src/conf.rs
```diff
@@ -143,18 +143,22 @@ pub struct Platform {
     /// - TODO: Document(and check) what does it actually mean on android. Transparent window?
     pub framebuffer_alpha: bool,
 
-    /// When using Wayland, this controls whether to draw the default window decorations.
+    /// On Wayland, the decorations are either drawn by the server or via `libdecor`. If neither is
+    /// available then no decorations will be drawn.
+    #[deprecated]
     pub wayland_use_fallback_decorations: bool,
```
not-fl3, line 149: I would keep this configurable. Maybe as an enum with options for no csd, ugly csd and libdecor?

not-fl3:
Wow, what a PR, great work! I do not have much to add, looks super good to me.

This is correct, this code was basically maintained by @narodnik, so I would wait for their feedback too.

src/conf.rs
not-fl3: I would still keep it configurable. Boolean worked when there was just two options: SSD only or SSD + CSD fallback, but now we got options! 

Said this, if you think its too much code to maintain - I am okay to just deprecating the flag and doing SSD/libdecor/no decorations automatically.

not-fl3: I do not think borderless should affect a normal ssd/csd fallback logic. For SSD/CSD selecteion I would either have an enum or fully automated, no configuration mechanism.

not-fl3: Thanks! What a PR! :heart:

not-fl3: Absolutely, lets do it!
I would personally prefer X11WithWaylandFallback, but with this PR I don't really see why not WaylandWithX11Fallback!

PR #536, opened 2025-03-10
MacOS: Fix gl_context null pointer dereference on startup

domenukk:
When using Macroquad, I started running into crashes on MacOS on startup recently. I'm not sure if it is due to a newer Rust version that started checking this, or due to a new MacOS version that started moving windows on startup.

In any case, ignoring a null `gl_context` fixed the issue, and everything else works as expected afterwards.


not-fl3:
> @not-fl3 could you maybe turn this into a minor release and update macroquad with the new version too?

sure, on it!

PR #537, opened 2025-03-14
Wayland fixes

bolphen:
A few fixes for the Wayland backend.

- The error handling is kinda important; without it, the compositor can decide to close the window with Miniquad busy looping in the background and consuming 100% CPU usage, which would be very bad for a clueless end user.
- Handle dpi scale change in runtime. This is where I ran into the error handling issue: if you have a window with size say 801x600, changing dpi to 2x would end up with a protocol error saying that 801 is not a multiple of 2. There's still some issue with the current implementation, as after a dpi scale change the first few frames would not be rendered correctly. But at least it's not crashing, and I think changing dpi really doesn't happen that frequently so no more Wayland's "every frame is perfect" :shrug:

### DIFF
--- a/src/native/linux_wayland.rs
+++ b/src/native/linux_wayland.rs
@@ -787,23 +787,16 @@ unsafe extern "C" fn relative_pointer_handle_relative_motion(
 }
 
 unsafe extern "C" fn output_handle_scale(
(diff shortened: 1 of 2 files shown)


not-fl3:
Thanks!!

PR #541, opened 2025-04-12
fix `RenderPass.resolves` texture leak

cyrgani:
Looks like it was forgotten to also free the resolve buffers and textures again when they were added to `RenderPassInternal`.
Fixes https://github.com/not-fl3/macroquad/issues/886.


not-fl3:
Thanks for the fix!

Issue #28, opened 2020-02-23
Window focus events


not-fl3:
OS window focus/unfocus events should be forwarded to EventHandler.

Usecases: pause game on window focus lost or release captured cursor on alt-tab to another window.

PR #552, opened 2025-05-03
fix and expand the support of the window_minimized_event

InnocentusLime:
## Why?

`macroquad` has a bug: not-fl3/macroquad#547.
There is a good fix we can implement, that has been suggested by @profan - the same issue, but in godot (godotengine/godot#28061). Essentially, this PR part 1 of the fix.

## Context

`miniquad` already has some ready-to-use handlers for a potential fix. There is `window_minimized_event`. Despite the name, on X Window system and WASM it gets fired when a window loses focus.

I decided not to change that and instead focus on the following:
* Adding a similar support for OSX (special thanks to @jonatino for the help!)
* Adding a similar support for Windows
* Fixing a related bug on WASM

## Bug Reproduction

The following application is sufficient to catch the bug

```rust
use macroquad::prelude::*;

#[macroquad::main("Catching Keys")]
async fn main() {
    loop {
        clear_background(WHITE);

        let col = if is_key_down(KeyCode::A) {
            GREEN
        } else {
            RED
        };
        draw_rectangle(0.0, 0.0, 32.0, 32.0, col);

        next_frame().await
    }
}
```

The steps are simple:
1. Run the app
2. Hold the A key
3. While still holding the key -- focus on a different window
4. Release the A key
5. Focus again on the window
6. The square remains green (because `is_key_down` is still returning `true`)

## Testing The Fix

1. Use the same setup
2. Add the following lines to your `Cargo.toml`
3. Repeat the same steps
4. The square will now turn red on step (6)!

```toml
[patch.crates-io]
# hosts the second part of the fix
macroquad = { git = "https://github.com/InnocentusLime/macroquad.git", branch="lab-repeat-keys" }
# This branch with the fix
miniquad = { git = "https://github.com/InnocentusLime/miniquad.git", branch="fix-window-minimized" }
```

Tests:
- [x] Works on Windows
- [x] Works on MacOS/OsX (tested by @birhburh)
- [x] Works on WASM
- [x] Works on X11 (tested by @profan)

## Technical Notes About the fixed `js/gl.js`

There were a few things that did not make the minimized events fire correctly:
1. ~~`lastFocus == hasFocus` caused the app being notified of focus only when it didn't change~~ (fixed in 8d7cb80)
2. the `visibilitychange`  listener had a mistake. Elements **do not** lose focus when the browser changes tabs of the browser window becomes inactive. Instead `document.visibilityState` becomes not equal to `"visible"`

### DIFF
--- a/js/gl.js
+++ b/js/gl.js
@@ -1365,7 +1365,9 @@ var importObject = {
 
             let lastFocus = document.hasFocus();
             var checkFocus = function () {
-                let hasFocus = document.hasFocus();
+                // The element doesn't loose focus when the user switches tabs.
+                // However, the document becomes invisible
+                let hasFocus = document.hasFocus() && document.visibilityState == "visible";
                 if (lastFocus != hasFocus) {
                     wasm_exports.focus(hasFocus);
                     lastFocus = hasFocus;
--- a/src/event.rs
+++ b/src/event.rs
@@ -206,9 +206,9 @@ pub trait EventHandler {
     fn raw_mouse_motion(&mut self, _dx: f32, _dy: f32) {}
 
     /// Window has been minimized
-    /// Right now is only implemented on Android, X11 and wasm,
+    /// Right now is only implemented on Android, Windows, OSX, X11 and wasm,
     /// On Andoid window_minimized_event is called on a Pause ndk callback
-    /// On X11 and wasm it will be called on focus change events.
+    /// On X11, OSX, Windows and wasm it will be called on focus change events.
     fn window_minimized_event(&mut self) {}
 
     /// Window has been restored
--- a/src/native/macos.rs
+++ b/src/native/macos.rs
@@ -324,6 +324,9 @@ pub fn define_cocoa_window_delegate() -> *const Class {
             // Startup: the gl_context has not yet been created.
             return;
         }
+        if let Some(event_handler) = payload.context() {
+            event_handler.window_minimized_event();
+        }
         unsafe {
             msg_send_![payload.gl_context, update];
         }
@@ -353,6 +356,18 @@ pub fn define_cocoa_window_delegate() -> *const Class {
                 let state: u64 = msg_send![payload.window, occlusionState];
                 payload.occluded = state & NSWindowOcclusionStateVisible == 0;
             }
+        }
+    }
+    extern "C" fn window_did_become_key(this: &Object, _: Sel, _: ObjcId) {
+        let payload = get_window_payload(this);
+        if let Some(event_handler) = payload.context() {
+            event_handler.window_restored_event();
+        }
+    }
+    extern "C" fn window_did_resign_key(this: &Object, _: Sel, _: ObjcId) {
+        let payload = get_window_payload(this);
+        if let Some(event_handler) = payload.context() {
+            event_handler.window_minimized_event();
         }
     }
 
@@ -389,6 +404,14 @@ pub fn define_cocoa_window_delegate() -> *const Class {
         decl.add_method(
             sel!(windowDidChangeOcclusionState:),
             window_did_change_occlusion_state as extern "C" fn(&Object, Sel, ObjcId),
+        );
+        decl.add_method(
+            sel!(windowDidBecomeKey:),
+            window_did_become_key as extern "C" fn(&Object, Sel, ObjcId),
+        );
+        decl.add_method(
+            sel!(windowDidResignKey:),
+            window_did_resign_key as extern "C" fn(&Object, Sel, ObjcId),
         );
     }
     // Store internal state as user data
--- a/src/native/windows.rs
+++ b/src/native/windows.rs
@@ -535,6 +535,13 @@ unsafe extern "system" fn win32_wndproc(
                 }
             }
         }
+        WM_ACTIVATE => {
+            if LOWORD(wparam as _) == WA_ACTIVE || LOWORD(wparam as _) == WA_CLICKACTIVE {
+                event_handler.window_restored_event();
+            } else {
+                event_handler.window_minimized_event();
+            }
+        }
         _ => {}
     }
 


jonatino:
I've also made similar changes here for reference:

Miniquad commits:
windows & wasm: https://github.com/VoltaSoftware/miniquad/commit/59549b66aba7368a06731c4b901846983dabd5c6
android & linux: https://github.com/VoltaSoftware/miniquad/commit/5a87f41b5d0537c9a312511a9b209405c6481342
macos: https://github.com/VoltaSoftware/miniquad/commit/3e3263384a4975b51e6eabbcb79b89bdd7aea0e0
ios: https://github.com/VoltaSoftware/miniquad/commit/9a8af70ea3f3f37fbbd86ee81e84aca9f496cb45

Macroquad commit:
https://github.com/VoltaSoftware/macroquad/commit/49972d9a16ce7b19cb08526ac16362138af70791?w=1#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759

InnocentusLime:
> I've also made similar changes here for reference:
> 
> Miniquad commits: windows & wasm: [VoltaSoftware@59549b6](https://github.com/VoltaSoftware/miniquad/commit/59549b66aba7368a06731c4b901846983dabd5c6) android & linux: [VoltaSoftware@5a87f41](https://github.com/VoltaSoftware/miniquad/commit/5a87f41b5d0537c9a312511a9b209405c6481342) macos: [VoltaSoftware@3e32633](https://github.com/VoltaSoftware/miniquad/commit/3e3263384a4975b51e6eabbcb79b89bdd7aea0e0) ios: [VoltaSoftware@9a8af70](https://github.com/VoltaSoftware/miniquad/commit/9a8af70ea3f3f37fbbd86ee81e84aca9f496cb45)
> 
> Macroquad commit: [VoltaSoftware/macroquad@49972d9?w=1#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759](https://github.com/VoltaSoftware/macroquad/commit/49972d9a16ce7b19cb08526ac16362138af70791?w=1#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759)

Thank you! Gives me more confidence, that this fix is on the right track :D

profan:
Verified on X11 with Ubuntu 24.04.2 LTS (with a somewhat esoteric window manager to be fair, running awesomewm, a tiling wm), not sure if we want/need someone running Wayland to try it as well or if it's fine with just X11?

Seems to be working as expected :)

InnocentusLime:
> Verified on X11 with Ubuntu 24.04.2 LTS (with a somewhat esoteric window manager to be fair, running awesomewm, a tiling wm), not sure if we want/need someone running Wayland to try it as well or if it's fine with just X11?
> 
> Seems to be working as expected :)
> 
> edit: had a friend try it under Hyprland (Wayland) on Arch as well and it seems to be working there too

Good question. I have no idea what policy macroquad has towards Wayland. Last time I checked the support was "unstable". So technically speaking, bugs for Wayland are to be expected are not required to be fixed by this PR.

That said, if Wayland does get fixed too -- good!

InnocentusLime:
@not-fl3 the PR has been tested on all declared platforms and seems to work! Any chance of it getting merged? :)

jonatino:
@InnocentusLime I think your osx changes are missing the event for when the window gets reactivated. 

You should use this instead because these are the actual activated/deactivated windows events for macos: https://github.com/VoltaSoftware/miniquad/commit/3e3263384a4975b51e6eabbcb79b89bdd7aea0e0

InnocentusLime:
> @InnocentusLime I think your osx changes are missing the event for when the window gets reactivated. 
> 
> You should use this instead because these are the actual activated/deactivated windows events for macos: https://github.com/VoltaSoftware/miniquad/commit/3e3263384a4975b51e6eabbcb79b89bdd7aea0e0
> 
> https://developer.apple.com/documentation/appkit/nswindowdelegate/windowdidbecomekey(_:)
> https://developer.apple.com/documentation/appkit/nswindowdelegate/windowdidresignkey(_:)

Interesting. I just did what Godot did in their PR. I suppose I can just alter the commit to use those instead. Thx! :)

InnocentusLime:
@jonatino 
Although, it looks like `DidMove` should trigger the minimized event too just in case. 
I will **add** the handlers instead.

narodnik:
Hey this is the same as this issue: https://github.com/not-fl3/miniquad/issues/517

While your fix should be applied probably, it doesn't fix the overall issue with the key APIs. For that we should copy what Qt does with making available the actual scancodes, as well as merging the char and various key events into a unified API.

InnocentusLime:
> Hey this is the same as this issue: #517
> 
> While your fix should be applied probably, it doesn't fix the overall issue with the key APIs. For that we should copy what Qt does with making available the actual scancodes, as well as merging the char and various key events into a unified API.

They API might be a bit confusing right now, but `KeyCode` is actually the key scancode. It is why not-fl3/macroquad#686 exists actually

profan:
@narodnik that's odd, considering there's this issue: https://github.com/not-fl3/macroquad/issues/686 which seems to imply all the key codes are related to the keyboard layout, so we have some awkward middleground right now then?

InnocentusLime:
> Just to weigh in on this: as I understand, what this PR does is to fire the `window_minimized_event` / `window_restored_event` upon focus lost / gain, and has nothing to do with keyboard input. Moreover these events are already being fired on Linux (both X11 and Wayland) before this PR; the PR just implements them on the other platforms (which btw is a good indication that the Wayland backend is actually pretty on a par with the other platforms).
> 
> There's also a related issue by Fedor himself #28 although I think we can all agree that this is basically `window_minimized` now.

You are absolutely correct. 
1. They key mapping is completely unrelated to this PR. 
2. Yes, this PR makes `window_minimized_event` and `window_restored_event` properly cross-platform

InnocentusLime:
@narodnik 
> Linux wayland does not use scancodes. It converts them to key syms, then converts the key syms to key codes. For tracking key state, you should use the raw key passed in (which it does), but this does not match the key codes used in miniquad. The miniquad KeyCode does not work for key repeat.

Apologies. I may not have a good grasp of how wayland works. 

Regardless, I believe `KeyCode` should remain related to physical key locations and nothing else. Otherwise it becomes difficult to use for gamedev purposes.

This PR is also in fact unrelated to #517, despite what you said. So I would suggest moving the keyboard processing related discussion there

not-fl3:
Oh, thanks for PR, missed it!

PR #553, opened 2025-05-03
Fixed bug with window focus event on wasm only sending focus = true event

jonatino:
Small bug with wasm focus logic.


not-fl3:
Good catch, thanks!

PR #556, opened 2025-05-16
graphics: (gl) error out when the uniform is not found

InnocentusLime:
* Error out when a uniform in `ShaderMeta` is absent
* Do not store `Option<GLint>` as uniform location

### DIFF
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -284,6 +284,9 @@ impl Display for ShaderType {
 
 #[derive(Clone, Debug)]
 pub enum ShaderError {
+    MissingUniform {
+        uniform: String,
+    },
     CompilationError {
         shader_type: ShaderType,
         error_message: String,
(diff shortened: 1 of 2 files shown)


not-fl3:
Will it prevent doing shadertoy-style extra, possibly unused uniforms? Say `_Time` that is always passed as the uniform, but almost never used?

InnocentusLime:
> Will it prevent doing shadertoy-style extra, possibly unused uniforms? Say `_Time` that is always passed as the uniform, but almost never used?

@not-fl3 
It will. However, I think it leads to more trade-offs than gains:

1. Yes, you will have silently ignored `_Time`, but it is almost the only case when a missing uniform is kind of okay
2. The silently passed `_Time` introduces an order-dependency among uniforms
3. The possibly-unused `_Time` just doesn't work on Metal backend, because we **have no is-used checks**
4. We have a constant `Some`-check in the GL backend
5. I do realise, that `macroquad` relies on that, but it can optionally provide or not provide `_Time` just fine instead of expecting `miniquad` to figure that out

To sum up, I think this part of `miniquad` should be simpler and this extra fallback should be part of the more complex applications.

InnocentusLime:
@not-fl3 
Please have another look. Optional uniforms do not work on Metal. 
I think it should either be supported on both ends or not guaranteed at all :/

not-fl3:
Sure, it should be supported on both ends! I believe we had a fix for metal?

InnocentusLime:
Not that I know of, no. The current miniquad just puts everything through as-is with no changes 
https://github.com/not-fl3/miniquad/blob/master/src/graphics/metal.rs#L1128-L1148

So if macroquad supplies `Time_` and nobody expects it, the shader will break

PR #563, opened 2025-06-21
android: Fixing two potential IndexOutOfBoundsException.

MohammedKHC0:
The first one is characters.length >= 0 It should be >= 1. The second one is acessing the first item of the primary clip while in theory it could be empty. Also btw ClipboardManager will never be null.


not-fl3:
Thanks for PR!

PR #566, opened 2025-07-13
allow minimum window size to be set for Windows OS

streak324:
The UI in my game is suboptimal when the window dimensions are too small. i wanted something to enforce that.


streak324:
closing to put this in a separate branch

PR #573, opened 2025-08-04
android: remove window param from SurfaceCreated msg. Calling create_native_window(surface) leaks a resource which is unused anyway since we have surfaceCreated. No reason to call this here

narodnik:


not-fl3:
Thank you!

PR #575, opened 2025-08-06
egl: add missing eglGetCurrentContext(), can be used to check whether the current EGL context is valid

narodnik:


narodnik:
bump

PR #578, opened 2025-08-16
android: allow calling window::schedule_update() from any thread with the blocking event loop active

narodnik:
Android with the non blocking event loop, uses mpsc rx.try_recv() repeatedly which drastically reduces the battery (revealed during my testing). The solution is then to use the blocking event loop, but due to the current design where requests are a separate queue, then calling schedule_update() will only work when the main queue has already been interrupted.

The reason calling schedule_update() inside events alone is insufficient is for async events like notifications such as connection status indicator which varies regardless of user action.

The solution then for android is to merge both these queues together. Since NativeDisplayData.native_requests() is a queue, we replace this with a callback instead. Then each native module can specify whichever logic they want to pass in requests. For android merging the queues together solves the issue.

For now I left Mac, Win, iOS and Linux untouched, but will gradually add these in since the goal is also to be able to use non blocking event loop on laptops and save battery as well.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -211,9 +211,17 @@ pub mod window {
     /// TODO: implement window focus events
     pub fn set_cursor_grab(grab: bool) {
         let d = native_display().lock().unwrap();
-        d.native_requests
-            .send(native::Request::SetCursorGrab(grab))
-            .unwrap();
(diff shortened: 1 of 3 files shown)


not-fl3:
🙏🙏🙏

PR #581, opened 2025-08-25
Use OpenGL 3.1 on Windows

jangler:
See https://github.com/not-fl3/macroquad/issues/257 - this change (downgrading the requested OpenGL version) fixes the issue without causing any adverse effects that I could notice, either on a system with integrated Intel graphics or dedicated Nvidia graphics.


not-fl3:
Thanks for PR!

PR #582, opened 2025-08-25
Fix iOS display scaling in high_dpi mode

vpzomtrrfrt:
Seems to fix not-fl3/macroquad#994 (At least for high_dpi=true, I don't know what the expected behaviour is for false)


not-fl3:
Thansk for PR!

PR #591, opened 2025-11-30
feat(windows): Add IME support for CJK text input

gqf2008:
- Fix IME initialization by moving ShowWindow() after WGL context creation
- Add window::set_ime_position(x, y) API for IME candidate window positioning
- Add window::set_ime_enabled(enabled) API to toggle IME for text input vs game controls
- Handle WM_IME_COMPOSITION, WM_IME_STARTCOMPOSITION, WM_IME_ENDCOMPOSITION messages
- Properly manage IME context on window focus changes
- Add placeholder implementations for Linux, macOS, iOS, and Android
- Add ime_test.rs example demonstrating IME input with text rendering

This enables proper Chinese/Japanese/Korean text input on Windows, which was previously broken due to IME not initializing correctly when the window was shown before the OpenGL context was created.


not-fl3:
Great work, thanks for PR!

PR #594, opened 2025-12-24
android: add platform config to set an FPS setting when `blocking_event_loop` is enabled

narodnik:
Currently in my app I have a task scheduling updates continuously but this introduces latency and uneeded cpu usage. Instead we can make the receiver optionally call .recv_timeout() inside the main event loop. In fact all native platforms in miniquad should use an FPS by default instead of continuous drawing.

```rust
            // For animations do periodic refresh every 40 ms
            self.refresh_task = Some(self.ex.spawn(async move {
                loop {
                    system::msleep(40).await;
                    miniquad::window::schedule_update();
                }
            }));
```

(on android you dont want to continuous draw() or else it will drain battery. In fact you dont want to do this on most platforms)

### DIFF
--- a/src/conf.rs
+++ b/src/conf.rs
@@ -152,6 +152,14 @@ pub struct Platform {
     /// [`schedule_update`]: super::window::schedule_update
     pub blocking_event_loop: bool,
 
+    /// When `blocking_event_loop` is enabled, we can set this value to periodically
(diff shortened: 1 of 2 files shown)


narodnik:
Should I rename `fps` to `target_fps` maybe?

Also if you prefer, we can put `sleep_interval_ms` instead of `fps`.

not-fl3:
Thanks for PR!

Issue #418, opened 2024-02-06
Metal incorrect output


daslastic:
## Summery
On MacOS, using Metal as the render API causes an incorrect output when defining or using:
- **FilterMode**: (i.e. sampler), 
- **Viewports**: apply_viewport
- **PrimitiveType**: (i.e. PrimitiveType::Lines)
- Everything else seems to be working but I can't be exactly sure until I've tested everything.

This is illustrated in this minimal example, allowing to easily choose the output OpenGl or Metal:
```
use miniquad::*;

#[repr(C)]
struct Vec2 {
    x: f32,
    y: f32,
}
#[repr(C)]
struct Vertex {
    pos: Vec2,
    uv: Vec2,
}

struct Stage {
    ctx: Box<dyn RenderingBackend>,

    pipeline: Pipeline,
    bindings: Bindings,
}

impl Stage {
    pub fn new() -> Stage {
        let mut ctx: Box<dyn RenderingBackend> = window::new_rendering_backend();

        #[rustfmt::skip]
        let vertices: [Vertex; 4] = [
            Vertex { pos : Vec2 { x: -0.5, y: -0.5 }, uv: Vec2 { x: 0., y: 0. } },
            Vertex { pos : Vec2 { x:  0.5, y: -0.5 }, uv: Vec2 { x: 1., y: 0. } },
            Vertex { pos : Vec2 { x:  0.5, y:  0.5 }, uv: Vec2 { x: 1., y: 1. } },
            Vertex { pos : Vec2 { x: -0.5, y:  0.5 }, uv: Vec2 { x: 0., y: 1. } },
        ];
        let vertex_buffer = ctx.new_buffer(
            BufferType::VertexBuffer,
            BufferUsage::Immutable,
            BufferSource::slice(&vertices),
        );

        let indices: [u16; 6] = [0, 1, 2, 0, 2, 3];
        let index_buffer = ctx.new_buffer(
            BufferType::IndexBuffer,
            BufferUsage::Immutable,
            BufferSource::slice(&indices),
        );

        let check_bytes: &[u8] = &[
            172, 50, 50, 255, 0, 0, 0, 0, 172, 50, 50, 255, 0, 0, 0, 0, 172, 50, 50, 255, 0, 0, 0,
            0, 172, 50, 50, 255, 0, 0, 0, 0, 172, 50, 50, 255, 0, 0, 0, 0, 172, 50, 50, 255, 0, 0,
[...]

PR #611, opened 2026-03-27
Improved Metal support

Pewnack:
This is a fix for issue #418 
- When using Metal, primitive type was hardcoded to Triangles. Now the primitive type is taken from the pipeline parameters.
- Setting the view port was not supported in Metal. Implemented apply_viewport.


Pewnack:
This might have been a bit of a hasty PR. I'm closing it.

Issue #561, opened 2025-06-08
miniquad calls macOS specific API when building for iOS


letrunghieuvina:
Hi,
I created a simple app to test only, it just draws a rectangle on screen. I built a static library with "--target aarch64-aaple-ios-sim", then I created an XCode project to use the app, XCode spat out this error when building

Undefined symbols for architecture arm64:
  "_CGDisplayPixelsHigh", referenced from:
      miniquad::native:🍎:apple_util::bottom_left_to_top_left::h4dc20aa8ec8d1c1d in liblearnmq.a[108]

By searching, it says that CDDisplayPixelsHigh only available on macOS.

This is the full log:

Undefined symbols for architecture arm64:
  “_CGDisplayPixelsHigh”, referenced from:
      miniquad::native::apple::apple_util::bottom_left_to_top_left::h4dc20aa8ec8d1c1d in liblearnmq.a[108](miniquad-7314d62e1368c346.miniquad.eaa95033056c3ec2-cgu.10.rcgu.o)
  “_CGMainDisplayID”, referenced from:
      miniquad::native::apple::apple_util::bottom_left_to_top_left::h4dc20aa8ec8d1c1d in liblearnmq.a[108](miniquad-7314d62e1368c346.miniquad.eaa95033056c3ec2-cgu.10.rcgu.o)
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)

PR #617, opened 2026-04-04
Some CoreGraphics functions are only available on macOS

Pewnack:
This is a fix for #561 

Apple's CoreGraphics framework functions CGDisplayPixelsHigh and CGMainDisplayID are only available on macOS. They do not exist on iOS.

If you compile your miniquad code as a static library and link it to your iOS App, the build on Xcode fails with the following errors:

Undefined symbol: _CGDisplayPixelsHigh
Undefined symbol: _CGMainDisplayID
Linker command failed with exit code 1 (use -v to see invocation)

Modified framework.rs to make sure, they're only included on macOS.

### DIFF
--- a/src/native/apple/frameworks.rs
+++ b/src/native/apple/frameworks.rs
@@ -188,8 +188,6 @@ extern "C" {
         window_id: u32,
         imageoptions: u32,
     ) -> ObjcId;
-    pub fn CGMainDisplayID() -> u32;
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks for PR! 

I very much appreaciate your work, I do not have any Mac hardware, so without your PRs Mac bugs would never be fixed :)

Pewnack:
Thanks for merging!

Issue #568, opened 2025-07-24
MacOS doesn't have file dropping implemented


Bonsix:
So, while I was working with file dropping I noticed that MacOS doesn't have it implemented. WASM, Linux and Windows all have it, only MacOS doesn't from non-mobile targets.

PR #619, opened 2026-04-06
Implemented files_dropped_event for macOS

Pewnack:
This is a fix for issue #568 

Implemented the files_dropped_event for macOS. It's only triggered for files, not directories. Also returns the dropped_file_bytes same as for linux_x11.

On macOS this only required the implementation of two events :
- draggingEntered - Triggered when drag 'n drop targets enter the Window and decides if you are going to accept it. Here we check if we are actually dragging file URLs. If not, the drag is not accepted. File URLs also include directories though.
- performDragOperation - Triggered when the drop takes place. Here we explicitly check if it is files we are dropping. Directories are skipped and will not trigger the files_dropped event.

We also need to tell the view to handle drag 'n drop with a call to registerForDraggedTypes. Without this, the drag 'n drop events are not triggered at all.

When files are dropped we read its bytes for dropped_file_bytes with std::fs::read. Mimicked this from linux_x11.rs.

Tested it on macOS Tahoe 26.4 (currently the latest macOS version) and on the older macOS Monterey 12.7.6.

### DIFF
--- a/src/native/macos.rs
+++ b/src/native/macos.rs
@@ -601,6 +601,75 @@ unsafe fn view_base_decl(decl: &mut ClassDecl) {
         }
     }
 
+    extern "C" fn dragging_entered(_this: &Object, _sel: Sel, sender: ObjcId) -> NSDragOperation {
+        unsafe {
+            let pasteboard: ObjcId = msg_send![sender, draggingPasteboard];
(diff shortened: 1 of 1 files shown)


Pewnack:
Thanks for merging!

PR #610, opened 2026-03-22
Data returned from Xlib/XGetEventData is not guaranteed to be aligned

Pewnack:
Fixes #394 
Data returned from Xlib is not guaranteed to be aligned. This could cause a "misaligned pointer deference" panic when getting the mouse deltas. The mouse deltas are two doubles, but the buffer in which they are stored might not be 8 byte aligned. This apparently could happen quite easily on a Linux 32 bit system.

Tested it on a 32bit Raspberry PI. Also tested this on a Ubuntu 64bit, just to see It didn't break anything there.

PR #621, opened 2026-04-11
native/linux_x11: Fix dy value in raw_mouse_motion

Pewnack:
On X11, event raw_mouse_motion returned incorrect dy value. This was introduced by #610. After reading the mouse x delta data pointer was only advanced by 1 byte instead of 8 bytes.


not-fl3:
Thanks for PR!

Pewnack:
Thanks for merging!

PR #623, opened 2026-04-18
native/macos: Add frame pacing for smooth motion

### LINKED ISSUE #559: Unstable framerate om macOS
I've been using macroquad which in turn uses miniquad version 0.4.0 with a stable framerate. I moved to using miniquad directly using 0.4.8 and noticed the framerate wasn't as stable (motion looked slightly janky). Compiling and running version 0.4.5 of the quad example works well, but at 0.4.6 the framerate is unstable again.

I measure the framerate difference by using `std::time::Instant`. Using `v0.4.5`, the framerate is stable at `8.3333ms`. Starting with version `v0.4.6`

Delta time in the quad example:

Version 0.4.5:
```
Frame time: 8.365 ms
Frame time: 8.302 ms
Frame time: 8.333 ms
Frame time: 8.308 ms
Frame time: 8.364 ms
Frame time: 8.289 ms
Frame time: 8.377 ms
Frame time: 8.295 ms
Frame time: 8.303 ms
Frame time: 8.340 ms
Frame time: 8.425 ms
Frame time: 8.225 ms
Frame time: 8.321 ms
```

Version 0.4.6:
```
Frame time: 8.397 ms
Frame time: 8.264 ms
Frame time: 8.313 ms
Frame time: 9.048 ms
Frame time: 1.930 ms
Frame time: 14.329 ms
Frame time: 2.098 ms
Frame time: 14.644 ms
Frame time: 2.510 ms
Frame time: 14.082 ms
Frame time: 2.165 ms
Frame time: 14.625 ms
Frame time: 2.841 ms
Frame time: 13.677 ms
Frame time: 2.353 ms
Frame time: 14.517 ms
Frame time: 2.453 ms
Frame time: 13.699 m
```

Sometimes the framerate even exceeds 30ms. All this makes the motion on screen look a little bit janky.

This is for a M2 macbook with "Promotion", e.g adaptive framerate display. I have observed the janky framerate on another macbook (without promotion, 60Hz display), and the jankyness is still there, but I don't have the delta time numbers for that machine.

I'm sticking with an older version for now, but I guess it's something that might need some attention.

Pewnack:
Fixes #559 

On macOS, if you are using OpenGL,  motion stutters. Motion is not smooth/fluent. You can see it immediately when you run
```
cargo run --example quad
```
The problem is that there isn't any real frame pacing in the current main loop. The main loop pushes out a new frame each iteration with flushBuffer and relies on VSYNC (swap_interval = 1) for pacing. But that's not enough on macOS. flushBuffer gives us the new frame, but it's up to the display server (Quartz Compositor) to decide when to use it.

This change adds a frame pacer using macOS's CVDisplayLink callback.  Basically what this change does is this:
- Adds a flag frame_ready that tells us when to start rendering the next frame.
- CVDisplayLink callback sets the flag to true
- The main loop wait for frame_ready to become true and continues with the next iteration.

There might be other ways to do it. But I wanted to keep the main loop intact as much as possible.

With this frame pacer motion is now super smooth :smile:

Before version v0.4.6 it was still reasonably smooth. That's because rendering was done differently. In v.0.4.5 each iteration of the main loop called setNeedsDisaply:YES on the view to trigger a redraw. This marks the view as 'dirty' signaling the display server to update it. The redraw performed a flushBuffer with VSYNC.

### DIFF
--- a/src/native/apple/frameworks.rs
+++ b/src/native/apple/frameworks.rs
@@ -238,6 +238,65 @@ extern "C" {
     pub fn MTLCopyAllDevices() -> ObjcId; //TODO: Array
 }
 
+#[cfg(target_os = "macos")]
+#[repr(C)]
+#[derive(Clone, Copy, Debug)]
+pub struct CVTime {
+    pub time_value: i64,
+    pub time_scale: i32,
+    pub flags: i32,
+    pub epoch: i64,
+}
+
+#[cfg(target_os = "macos")]
+pub type CVDisplayLinkRef = *mut c_void;
+
+#[cfg(target_os = "macos")]
+pub type CVOptionFlags = u64;
+
+#[cfg(target_os = "macos")]
+#[repr(C)]
+#[derive(Clone, Copy, Debug)]
+pub struct CVTimeStamp {
+    pub version: u32,
+    pub video_time_scale: i32,
+    pub video_time: i64,
+    pub host_time: u64,
+    pub rate_scalar: f64,
+    pub video_refresh_period: i64,
+    pub smpte_time: i64,
+    pub flags: u64,
+    pub reserved: u64,
+}
+
+#[cfg(target_os = "macos")]
+pub type CVDisplayLinkOutputCallback = extern "C" fn(
+    display_link: CVDisplayLinkRef,
+    now: *const CVTimeStamp,
+    output_time: *const CVTimeStamp,
+    flags_in: CVOptionFlags,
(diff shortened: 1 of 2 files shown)


not-fl3:
One day I'll get myself a mac and will familiarize myself with all this apple magic 🫣

Thanks for PR!

Pewnack:
Learned a lot about the 'apple magic' myself while making this PR 😄 Thanks for the merge!

PR #624, opened 2026-04-19
native/linux_wayland: Respect window_resizable

bolphen:
Sorry didn't test under gnome last time; we are using libdecor for the window decoration, and to disable window resizing we have to go through libdecor as well.


not-fl3:
Thanks for PR!

PR #625, opened 2026-04-25
native/macos: Added support for CursorIcon::NESWResize and CursorIcon::NWSEResize

### LINKED ISSUE #1030: CursorIcon::NESWResize and CursorIcon::NWSEResize work on Macos with set_mouse_cursor
Simply put,
```rs
set_mouse_cursor(CursorIcon::NESWResize);
``` 
and
```rs
set_mouse_cursor(CursorIcon::NWSEResize);
``` 
don't work on MacOS (and possibly linux). This could be due to an old version of GLFW though I'm not sure if macroquad/miniquad uses GLFW.

Pewnack:
On macOS cursor icons CursorIcon::NESWResize and CursorIcon::NWSEResize were not supported. Added support for it. 
These cursor icons are not officially supported, but there is an undocumented API for it.

The code for it was basically already there. In apple_util.rs there already is a function load_undocumented_cursor. This function was not entirely correct though and caused a panic. It needed only a small change to make it work. It now tries to set the undocumented cursor and falls back to the default cursor if it doesn't exist.

Cursor icons can be tested with miniquad example program mouse_cursor:
```bash
cargo run --example mouse_cursor
```

Press keys q to set NESWResize or w to set the NWSEResize icon.

Closes not-fl3/macroquad#1030

### DIFF
--- a/src/native/apple/apple_util.rs
+++ b/src/native/apple/apple_util.rs
@@ -43,8 +43,12 @@ pub fn load_undocumented_cursor(cursor_name: &str) -> ObjcId {
     unsafe {
         let class = class!(NSCursor);
         let sel = Sel::register(cursor_name);
-        let sel: ObjcId = msg_send![class, respondsToSelector: sel];
-        let id: ObjcId = msg_send![class, performSelector: sel];
+        let has_selector: BOOL = msg_send![class, respondsToSelector: sel];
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks for PR!

Pewnack:
Thanks for the merge!

PR #626, opened 2026-04-27
native/macos: Fix key up events when command key is pressed

### LINKED ISSUE #980: `is_key_down` gets "stuck" when chording with Command on macOS
## Reproduction
```rust
use macroquad::prelude::*;

#[macroquad::main("MyGame")]
async fn main() {
    loop {
        if is_key_down(KeyCode::D) {
            println!("D is down");
        }

        next_frame().await
    }
}
```

When the program is running, press `CMD+d` as a chord. Until D is pressed again, the print will repeatedly fire. (There is nothing special about "D", it's just the arbitrary key I chose in my reproduction. Other modifier keys work fine).

## Versions

```toml
[dependencies]
macroquad = "0.4.14"
```

```
$ rustc --version
rustc 1.87.0 (17067e9ac 2025-05-09)
```

Pewnack:
On macOS when the command key (super key) is pressed all KeyUp events are swallowed until the command key is released again. So when you press COMMAND + <key> you get a KeyDown event for both COMMAND and <key>, but when you release both keys again you only get a KeyUp event for the COMMAND key. 

This is fixed by faking these events ourselves. In the main loop (the run function) we already trap the raw event. We use this event to generate KeyUp events.

Closes not-fl3/macroquad#980

### DIFF
--- a/src/native/macos.rs
+++ b/src/native/macos.rs
@@ -131,6 +131,21 @@ impl MacosDisplay {
 
         Some(event_handler)
     }
+
+    // When the command key is pressed (super key), macOS swallows all key up events.
+    fn generate_swallowed_key_events(&mut self, event: ObjcId) {
+        let event_type: u64 = unsafe { msg_send![event, type] };
+        if event_type == NSEventType::NSKeyUp as u64 {
+            let mods = get_event_key_modifier(event);
+            if mods.logo {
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks for PR!

Pewnack:
Thanks for the merge!

PR #628, opened 2026-05-03
Fix link name for System in /native/apple/framework.rs

Roboticist01:
Previous link name was "system," which doesn't exist, but may resolve properly with particularly forgiving linkers or case-insensitive filesystems. Actual name is "System." This is basically a duplicate of #411, but based a newer commit of Miniquad that no longer seems to include CoreMIDI.


not-fl3:
Thanks for PR!

PR #629, opened 2026-05-18
Fix macOS cursor grab warp position

73v3:
Fixes incorrect cursor warp position on macOS when cursor grab is enabled.

Related Macroquad issue: https://github.com/not-fl3/macroquad/issues/858

## Problem

On macOS, enabling cursor grab can warp the cursor to an incorrect vertical position, often near the Dock/taskbar, instead of the center of the application content area.

The previous implementation computed the warp point from the outer `NSWindow` frame and then passed that through `transform_mouse_point()`. This mixes AppKit window/frame coordinates with CoreGraphics cursor-warp coordinates and can produce an incorrect Y position, especially for non-square windows.

## Fix

Use the actual content view bounds, convert the view-center point through AppKit using `convertPoint:toView:` and `convertRectToScreen:`, then flip the Y coordinate for CoreGraphics before calling `CGWarpMouseCursorPosition`.

This avoids manually combining outer window frame coordinates, titlebar/border offsets, scale factors, and screen-space conversion.

## Tested

Tested locally on macOS Apple Silicon with a Macroquad-based app using `set_cursor_grab(true)`.

Before this patch:
- enabling cursor grab warped the OS cursor near the Dock/taskbar;
- first-person/camera-style mouse control was broken.

After this patch:
- enabling cursor grab no longer moves the cursor to the Dock/taskbar area;
- the cursor remains correctly positioned in the game view;
- relative mouse/camera movement works correctly.

### DIFF
--- a/src/native/macos.rs
+++ b/src/native/macos.rs
@@ -144,12 +144,46 @@ impl MacosDisplay {
 
     fn move_mouse_inside_window(&self, _window: *mut Object) {
         unsafe {
-            let frame: NSRect = msg_send![self.window, frame];
-            let origin = self.transform_mouse_point(&frame.origin);
-            let point = NSPoint {
-                x: (origin.0 as f64) + (frame.size.width / 2.0),
-                y: (origin.1 as f64) + (frame.size.height / 2.0),
+            // Use the drawable/content view, not the outer NSWindow frame.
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks for PR!

PR #630, opened 2026-05-19
android: map the rest of the hardware-keyboard keycodes

benface:
## Summary

The Android keycode translation table in `src/native/android/keycodes.rs` covered
letters, digits, arrows, Enter, Tab, Space, Backspace, Shift / Alt, and the two
punctuation keys (comma, period). Most other keys an Android hardware keyboard
sends arrived as `KeyCode::Unknown`, which means tablet-with-keyboard, ChromeOS,
and Stage-Manager-iPad-with-a-Mac-keyboard setups couldn't reach the host app
with `Escape`, `Ctrl`, `Meta`/`Super`, function keys, the numpad, `Insert` /
`Delete` / `PageUp` / `PageDown` / `Home` / `End`, or several common punctuation
keys, even though `KeyCode` variants for all of them already exist.

The Ctrl/Super issue is particularly visible: #403 wired `KeyMods` tracking for
those modifiers, but as long as `KEYCODE_CTRL_LEFT` / `KEYCODE_META_LEFT`
translate to `KeyCode::Unknown`, the modifier state stays at default.

This PR adds entries for the keys that have an obvious one-to-one mapping with
an existing `KeyCode` variant. Specifically:

- **Modifiers**: `Ctrl L/R`, `Meta L/R` (Cmd on Apple keyboards), `CapsLock`,
  `NumLock`, `ScrollLock`.
- **Edit cluster**: `Escape`, `Insert`, `Delete` (`KEYCODE_FORWARD_DEL`,
  distinct from the existing `KEYCODE_DEL` → `Backspace`), `MoveHome`,
  `MoveEnd`, `PageUp`, `PageDown`, `PrintScreen` (`SYSRQ`), `Pause` (`BREAK`).
- **Punctuation**: `=`, `[`, `]`, `\`, `;`, `/`, backtick. (Apostrophe and minus
  have no matching variant in `KeyCode` so I left them out.)
- **Menu** (`KEYCODE_MENU`, the soft-menu / context-menu key on tablets).
- **`F1`–`F12`**.
- **Numpad**: `0`–`9` plus the four arithmetic ops, dot, `Enter`, and `=`.

The existing `KEYCODE_HOME` (`0x03`) entry that maps the device home button to
`KeyCode::Home` is kept — that keycode doesn't reach apps in practice, and the
new `KEYCODE_MOVE_HOME` (`0x7a`) maps to the same variant for the cursor-home
case, which is the one apps usually want.

I added trailing `// KEYCODE_*` comments on the previously-mapped lines so the
table is self-explanatory at a glance.

## Test plan

- [x] `cargo check --target aarch64-linux-android` — clean.
- [x] Built a downstream app (CHOMP) against this branch with a
      `[patch.crates-io] miniquad = { path = "…" }` override, deployed to a
      Pixel-Tablet AVD, and confirmed via `adb shell input keyevent
      KEYCODE_ESCAPE` that the host app's `is_key_pressed(KeyCode::Escape)`
      check now fires.
- [ ] Reviewers' usual sanity-build matrix.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/native/android/keycodes.rs
+++ b/src/native/android/keycodes.rs
@@ -1,28 +1,31 @@
 use crate::event::KeyCode;
 
+/// Translate an Android `AKEYCODE_*` value into miniquad's
+/// [`KeyCode`]. Keep entries ordered by numeric value; the comments
+/// reference the `KEYCODE_*` symbol from Android's `KeyEvent.java`
+/// so each line is self-explanatory at a glance.
 pub fn translate_keycode(keycode: u32) -> KeyCode {
-    // same as GLFW
     match keycode {
-        0x01 => KeyCode::Left,
-        0x02 => KeyCode::Right,
-        0x03 => KeyCode::Home,
-        0x04 => KeyCode::Back,
-        0x07 => KeyCode::Key0,
-        0x08 => KeyCode::Key1,
-        0x09 => KeyCode::Key2,
-        0x0a => KeyCode::Key3,
-        0x0b => KeyCode::Key4,
-        0x0c => KeyCode::Key5,
-        0x0d => KeyCode::Key6,
-        0x0e => KeyCode::Key7,
-        0x0f => KeyCode::Key8,
-        0x10 => KeyCode::Key9,
-        0x13 => KeyCode::Up,
-        0x14 => KeyCode::Down,
-        0x15 => KeyCode::Left,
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks for PR!

PR #631, opened 2026-05-22
fix: window resize flicker on Windows

lofcz:
During resize, Windows changes the client area before miniquad’s cached screen_width/screen_height are refreshed. This causes one frame to be rendered with stale viewport/projection.

Before:
```rust
update()
draw()
SwapBuffers()
update_dimensions()
resize_event()
```

After:
```rust
update_dimensions()
resize_event()
update()
draw()
SwapBuffers()
```


not-fl3:
Thanks for PR!

PR #634, opened 2026-06-13
ios: support iOS 27 UIScene lifecycle (TN3187)

benface:
## Problem

The iOS 27 SDK enforces the UIScene lifecycle for window management (per Apple's [TN3187](https://developer.apple.com/documentation/technotes/tn3187-migrating-to-the-uikit-scene-based-life-cycle)). Two consequences for apps built with the new SDK:

1. Apps that don't declare `UIApplicationSceneManifest` in `Info.plist` refuse to launch with the error _"UIScene life cycle is required for apps built with this SDK"_.
2. A `UIWindow` created via `initWithFrame:` from `application:didFinishLaunchingWithOptions:` is no longer auto-attached to a scene, so it stays orphan and invisible. The user sees a black screen after the splash; the GL/Metal view is never visible.

miniquad's current iOS path hits both. Note that **already-shipped apps run fine on iOS 27** — the enforcement is SDK-bound, not runtime-bound. The bug only appears when rebuilding against Xcode 27 / the iOS 27 SDK.

## Fix

This patch:

1. **Defers `UIWindow` creation** from `did_finish_launching_with_options` to a `UISceneWillConnectNotification` observer (`scene_will_connect`). The observer creates the window via `initWithWindowScene:` with the scene iOS provides, then adopts the view + view controller that `didFinishLaunching` had already set up. The view + view_ctrl are stashed in `thread_local` storage between the two callbacks (both run on the main thread).

2. **Observes `UISceneDidActivateNotification`** to re-send `Message::Resume`. Without this, iOS 27 fires a spurious `applicationWillResignActive:` during the scene-attach handoff (legacy lifecycle compat path) which sends `Message::Pause` and blocks the render thread on `rx.recv()`, with no follow-up `applicationDidBecomeActive:` to wake it back up. The symptom is exactly one frame of work happening (whatever ran synchronously up to the first `next_frame().await`), then nothing.

The patch is iOS-only and additive — pre-iOS-13 paths aren't touched, and `UISceneWillConnectNotification` has been available since iOS 13.

## Caller responsibilities

Apps using a miniquad with this patch **must declare `UIApplicationSceneManifest` in `Info.plist`** with at least one scene configuration, including a `UISceneDelegateClassName` pointing to either a stub `UIWindowSceneDelegate`-conforming class or the platform-provided default. Without the manifest, iOS 27 still refuses to launch the app — that part of the enforcement is unrelated to the patch.

A minimal stub delegate is enough; this patch handles all window/view wiring itself, so the scene delegate doesn't need any methods implemented:

```objc
@interface MySceneDelegate : UIResponder <UIWindowSceneDelegate>
@property (strong, nonatomic) UIWindow *window;
@end

@implementation MySceneDelegate
@end
```

## Tested on

iPhone running iOS 27 beta. Verified: launches past splash, window is visible, render thread runs continuously, asset loading completes, game renders + responds to input. Tested against the unpatched `miniquad@0.4.10` baseline (black screen / hang after first frame) for comparison.

## Why not a separate version bump

Happy to bump if preferred — this branch is built on top of the last commit at `version = "0.4.10"` on master to keep the diff minimal. Patch is ~80 lines net.

### DIFF
--- a/src/native/ios.rs
+++ b/src/native/ios.rs
@@ -24,6 +24,16 @@ use {
     },
 };
 
+// iOS 27 / TN3187: windows must be created via `initWithWindowScene:`
+// or they stay invisible. Window creation defers to the
+// `UISceneWillConnectNotification` observer below; both that and
+// `did_finish_launching_with_options` run on the main thread, so
+// `thread_local` storage is sound.
+thread_local! {
+    static PENDING_SCENE_VIEW: RefCell<Option<ObjcId>> = RefCell::new(None);
+    static PENDING_SCENE_VIEW_CTRL: RefCell<Option<ObjcId>> = RefCell::new(None);
+}
+
 struct MainThreadState {
     quit: bool,
     paused: bool,
@@ -493,7 +503,7 @@ pub fn define_app_delegate() -> *const Class {
     let mut decl = ClassDecl::new("NSAppDelegate", superclass).unwrap();
 
     extern "C" fn did_finish_launching_with_options(
-        _: &Object,
+        delegate_self: &Object,
         _: Sel,
         _: ObjcId,
         _: ObjcId,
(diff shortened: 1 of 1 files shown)


not-fl3:
Amazing work! 

As for the version bump - for sure all your updates deserve a new version, will do when I finish with the review.

PR #638, opened 2026-06-13
metal: implement multisample resolve textures

benface:
## Problem

`new_render_pass_mrt` accepts an optional `resolve_img` slice (per-color-attachment resolve target) but unconditionally panics with `unimplemented!` when any caller passes one. macroquad's MSAA render targets do pass them, so any app with `sample_count > 1` aborts on the first render-target construction:

```
PanicHookInfo { ... location: Location { file: ".../graphics/metal.rs", line: 568 } }
thread caused non-unwinding panic. aborting.
```

## Fix

Wire each `resolve_img[i]` to the matching color attachment's `resolveTexture` and flip the store action to `MultisampleResolve`. The non-resolve path is unchanged (`StoreAction = Store`).

Asserts the resolve slice length matches the color slice — the per-index pairing is implicit in the API and a length mismatch would silently drop or out-of-bounds otherwise.

## Tested on

iPhone 17 simulator on iOS 27 beta — a macroquad app with `sample_count: 4` (creating MSAA render targets) no longer aborts on first construction.

### DIFF
--- a/src/graphics/metal.rs
+++ b/src/graphics/metal.rs
@@ -539,8 +539,12 @@ impl RenderingBackend for MetalContext {
         resolve_img: Option<&[TextureId]>,
         depth_img: Option<TextureId>,
     ) -> RenderPass {
-        if resolve_img.is_some() {
-            unimplemented!("resolve textures are not yet implemented on metal");
+        if let Some(resolve_img) = resolve_img {
+            assert_eq!(
+                resolve_img.len(),
+                color_img.len(),
+                "resolve_img must have one entry per color attachment"
+            );
         }
         unsafe {
             let render_pass_desc =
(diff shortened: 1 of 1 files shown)


not-fl3:
Thanks for PR!

benface:
@not-fl3 – You're welcome. 😁 Did you mean to merge this? hehe

not-fl3:
oh 😅

benface:
Thank you @not-fl3 <3 By the way, did you get my emails?

not-fl3:
@benface Hmm, no, I don't think so. 

not.fl3@gmail.com / fedor@elvario.mx

benface:
@not-fl3 – Weird!

CleanShot 2026-07-25 at 13 37 46@2x

not-fl3:
Just triple checked, nothing in the inbox, nothing in spam - searched for anything with "miniquad" in the topic :crying_cat_face: 
No idea whats up with the gmail, I wonder how many emails am I missing...

Could you please forward it to fedor@elvario.mx? Or just send it again, maybe it will go through

benface:
Just sent it to that address @not-fl3, thanks so much for your time.

not-fl3:
Got it! :heart_hands:

PR #640, opened 2026-06-13
metal: honor Conf.sample_count for MSAA on iOS + matching pipelines

benface:
Three coupled changes that together light up MSAA on the iOS Metal backend.

## 1. `create_metal_view` calls `setSampleCount:` on `MTKView`

The `sample_count` argument was prefixed `_` and silently dropped, so `Conf.sample_count` had no effect on iOS Metal. macOS already does this.

## 2. `new_pipeline` matches the view's `sampleCount`

A pipeline's `sampleCount` must match every render-pass color attachment it draws to. Reading the view's value at pipeline-creation time keeps the main drawable and `render_target_msaa()` targets compatible with the same pipeline.

## 3. `new_texture` creates `D2Multisample` textures when `sample_count > 1`

For `RenderTarget` access with `sample_count > 1`, the color texture has to be a `D2Multisample` with matching `setSampleCount:` and `mipmapLevelCount = 1`, and can only carry `RenderTarget` usage (not `ShaderRead`/`ShaderWrite`). The single-sample resolve texture for the same target is created separately by the caller (e.g. macroquad's `render_target_ex`) and passed in via `new_render_pass_mrt`'s `resolve_img` slice — handled in #638.

## Constraint

Every render-pass color attachment must use the view's sample count. Mixing single-sample and multisample targets in one app would require per-target pipeline states, which Metal allows but miniquad's "one pipeline per material" model does not currently express. Callers that opt into `Conf.sample_count > 1` should use `render_target_msaa()` (or pass a matching `sample_count`) for every offscreen color target.

Builds on #634, #635, #636, #637, #638, #639.

### DIFF
--- a/src/graphics/metal.rs
+++ b/src/graphics/metal.rs
@@ -791,12 +791,9 @@ impl RenderingBackend for MetalContext {
                 }
                 msg_send_![descriptor, setStorageMode: MTLStorageMode::Private];
                 if params.sample_count > 1 {
-                    // MSAA render target. The matching single-sample
-                    // resolve texture is created with `sample_count = 1`
-                    // and passed alongside in `new_render_pass_mrt`'s
-                    // `resolve_img` slice. Multisample textures can't be
-                    // shader-sampled directly (they're sampled via the
-                    // resolve pass) and can't have mipmaps.
+                    // MSAA target — render-only, no mipmaps, no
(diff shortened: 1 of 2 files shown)


benface:
@not-fl3 – I just rebased this, since it had conflicts. Ready for your review. 🙏

not-fl3:
Thanks for PR!

PR #641, opened 2026-06-13
metal: collapse consecutive same-target passes into one encoder

benface:
Higher-level engines like macroquad issue one `begin_pass` / `end_render_pass` pair per draw call so they can stream batched draws to whatever the active camera's render target is. Each pair was becoming its own `MTLRenderCommandEncoder`, which on tile-based GPUs forces a full color-attachment store + (for MSAA) resolve on every draw — heavy memory bandwidth that visibly tanks real-device frame rate at retina resolution with `Conf.sample_count > 1`.

OpenGL absorbs this pattern transparently because re-binding an FBO doesn't trigger a resolve; Metal's per-encoder load/store actions make the cost explicit, and minimizing encoder count is a standard Metal best practice for exactly this reason.

This PR defers `endEncoding` until the next `begin_pass` actually requires a new encoder (different target, `Clear` action, or `commit_frame`). If the next `begin_pass` continues the same target with `PassAction::Nothing`, it keeps encoding into the existing encoder so the whole sequence collapses into one Metal pass with a single store/resolve at the real end. Each existing draw call's `apply_pipeline` / `apply_bindings` / `apply_uniforms` already re-sends its full state, so reusing the encoder is correctness-safe.

Tested on iPhone running iOS 27 with macroquad's per-draw-call begin/end pattern at 4× MSAA — drops the per-frame encoder count from the draw-call count down to one per render target, eliminating real-device frame lag.

### DIFF
--- a/src/graphics/metal.rs
+++ b/src/graphics/metal.rs
@@ -282,6 +282,10 @@ pub struct MetalContext {
     /// for `apply_scissor_rect`'s Y-flip + clamp.
     current_pass_width: u32,
     current_pass_height: u32,
+    // Bookkeeping for the deferred-end pass-merge in `begin_pass`
+    // / `end_render_pass`.
+    current_pass_target: Option<Option<RenderPass>>,
+    pending_end_encoder: bool,
     view: ObjcId,
     device: ObjcId,
     current_frame_index: usize,
@@ -375,6 +379,8 @@ impl MetalContext {
                 render_encoder: None,
                 current_pass_width: 0,
                 current_pass_height: 0,
+                current_pass_target: None,
+                pending_end_encoder: false,
                 view,
                 device,
                 buffers: vec![],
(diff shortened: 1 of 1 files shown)


not-fl3:
Uff thats a good one, happy to see IOS side of miniquad seeing some light! 

Thanks for PR!

benface:
Force-pushed: amended the commit to also call `really_end_encoder()` in `texture_generate_mipmaps`. Hit this on iOS — `MTLDebugCommandBuffer blitCommandEncoder` validates "encoding in progress" if a render encoder is still open on the same command buffer, which the deferred-end mechanism introduced in this PR makes possible. The blit encoder for mipmap generation was the missing audit site. Apologies for the miss.

benface:
Gentle bump @not-fl3 😇

not-fl3:
Sorry, missed it! Merged!

benface:
Don't be sorry @not-fl3! I'm really glad I can use the official version now instead of my fork. :D

PR #644, opened 2026-06-13
metal: grow per-buffer rotation Vec on demand instead of pre-allocating 30 slots

benface:
The Metal backend pre-allocates `BUFFERS_IN_ROTATION = 30` `MTLBuffer`s per logical `Buffer` to handle the within-frame `update; draw; update; draw` pattern — Metal's `update` doesn't synchronously claim ownership, so each update needs its own backing buffer to keep the previous draw's data alive. 30 was an arbitrary upper bound on per-frame multi-update count. For a non-trivial app this multiplies GPU buffer memory by ~30×; most of the slots are never touched.

Replace `Buffer.raw` with a `Vec<ObjcId>` initialised with a single backing buffer. `buffer_update` grows it lazily — when `next_value` exceeds `raw.len()`, allocate one more `MTLBuffer` with the cached `MTLResourceOptions` and push. After a few frames hit the app's actual per-frame max, growth stops. Apps that update each buffer once per frame carry a single backing buffer; apps that flush several times per frame on the same buffer grow that buffer's `raw` to match. The GL backend already does this transparently via driver-side buffer object orphaning behind `glBufferSubData`; this brings Metal in line with that behaviour.

## Reproduction

Surfaced by an Xcode GPU Workload capture on a 2D game with ~100 macroquad-binding slots: 1.5 GiB of buffer memory, of which only the first slot per binding was ever touched per frame. The over-allocated pool caused per-level lag as iOS GC'd around it. After the patch the same app sat at ~120 MiB of buffer memory and the lag disappeared.

`Buffer` loses `Copy` (Vec isn't Copy); verified no caller copies `Buffer` by value. `delete_buffer` already iterated `buffer.raw` and needs no change.

### DIFF
--- a/src/graphics/metal.rs
+++ b/src/graphics/metal.rs
@@ -191,20 +191,19 @@ fn roundup_ub_buffer(current_buffer: u64) -> u64 {
     ((current_buffer) + ((UNIFORM_BUFFER_ALIGN) - 1)) & !((UNIFORM_BUFFER_ALIGN) - 1)
 }
 
-// this scenario:
-// buffer.update(); draw(buffer); buffer.update(); draw(buffer);
-// is very problematic with metal's ownership model.
-// buffer.update() half-update buffer - it doesn't really send data to the GPU claiming ownership,
-// it postpone reading the CPU memory until drawcall will actually happen.
-// There is no way to flush the buffer and makes metal's "update" function to update GPU memory right away.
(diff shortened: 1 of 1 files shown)


not-fl3:
That was a lot of allocated buffers! Thansk for PR! 

Merging, but this made me think that we need examples/regressio_*.rs examples. For this case - with this `update() draw() update() draw()` chain.

PR #645, opened 2026-06-13
cfg-gate Android-only set_or_replace_display

benface:
The function is only called on Android (its own doc comment says so). Without the cfg gate, every other target (iOS, macOS, desktop, web) gets a \`dead_code\` warning on every build. Adding \`#[cfg(target_os = "android")]\` so it's only compiled on the platform that actually uses it.


not-fl3:
Done with the review! Really great job! Please ping me if I miss something. 
Your PRs made me very happy, I am glad someone is keeping an eye on IOS miniquad!

